PR #45, opened 2015-03-25
Fix enum field size (credit to @Arkshine)

Misiur:
Ref https://github.com/Zeex/pawn/issues/44


Zeex:
You mean squash the commits? I think It's fine, there will a merge commit anyway.

Issue #46, opened 2015-03-28
Fails to build on Debian amd64


pastcompute:
Also the instructions in the readme.txt don't accurately describe how to build from git - there are two sources directories (`amx`, `compiler`) which need separate cmake build directories

Various compiler errors. including:
```
pawn/source/amx/amx.c:493:5: error: enumerator value for ‘assert_static__’ is not an integer constant
```

PR #47, opened 2015-03-28
Fix to build on amd64 (fixes issue #46)

pastcompute:
And instructions on building from git


pastcompute:
Fixes #46

oscar-broman:
Cell size 64 is not compatible with plugins or even the SA-MP server.

pastcompute:
hmm - does this mean I cant try this out on amd64?
(Although if I find this project useful, I'll be targeting it at an embedded system, which means then I'll be trying to work out how to cross-compile the code runner for the embedded target whilst making the compiler still run on amd64... )

oscar-broman:
You could just compile it like this:
```
cmake . -DCMAKE_C_FLAGS=-m32 -DCMAKE_CXX_FLAGS=-m32
```

Zeex:
@pastcompute You might be better off with the official compiler from Compuphase, it has lots of bug fixes and some of the features of this compiler (though possibly in a slightly different form). They also released a new major version already a couple of years ago (4.0).

This compiler is more specialized on SA-MP, it may be not good for general use. In fact we don't use cell sizes other than 32 here as @oscar-broman noted. So the proposed changes would be of no use to us. Your contribution is appreciated but I won't merge this PR (sorry).

pastcompute:
No worries. I was just scouting around for some embedded scripting things to try out, and ended up here because the Compuphase repository was on google code which is going to get closed, I didnt realise this was different

Y-Less:
It would be interesting to see how compatible the two compilers are now.  As Zeex says, stick to the official compiler if you have total control over which version of the VM you use as well so you can keep the two in sync (and use the latest VM fixes as well).  There has been some back-and-forth between the official, SA:MP, and AMXModX compiler versions and features, but the latter two are quite specialised (and both stick to version 3 of the language).

This compiler is built on version 3.2.3664, with many fixes added or back-ported by Zeex, and some other features added (some of which were also adapted for the official compiler).  This link lists a lot of the official updates since the version we use (the VM is even older than the compiler):

http://www.compuphase.com/pawn/pawnhistory.htm

There are actually a couple of fixes there whose bugs we exploit in some ways.  This compiler supports the now official compile-time stringising and concatenation, but without the "..." operator that was added when that feature was back-ported.  Check these two links as well:

https://github.com/Zeex/pawn/wiki/What%27s-new
https://github.com/Zeex/pawn/wiki/Known-compiler-bugs

As I say, using the latest compiler and VM is probably the best bet, but on the other hand there is now a huge amount of documentation, knowledge, and libraries that are quite specific to this version or a range of close versions that could be reused.

Issue #43, opened 2015-03-24
Maximum array depth


Misiur:
Would it be possible to increase maximum array depth to 4<? Compiler flag would suffice. Example usage
```pawn
#define MAX_QUESTIONS 128
#define MAX_ANSWERS 4

enum E_TYPE {
    TYPE_ONE,
    TYPE_TWO
}

static const
     QuestionAnswers[E_TYPE][MAX_QUESTIONS][MAX_ANSWERS][64]
;
```

Zeex:
Well, I've found the constant responsible for the maximum number of array dimensions:

https://github.com/Zeex/pawn/blob/master/source/compiler/sc.h#L53

I'm not sure if it would be sufficient to change its value alone, but the code compiled.

Is it practical to set it to something bigger than 4?

Arkshine:
Be careful, I know SourceMod is using it and BAILOPAN said me it was kind of buggy.
Also, you might need this, which fixes some generation issue: https://hg.alliedmods.net/sourcemod-central/rev/ee044380947f
I don't know if there is others fixes around 4d.

Zeex:
I've changed the max. number of dimensions to 4 for the sake of experiment, however that seems not enough because 4d array initialization doesn't seem to be fully supported:

```pawn
#include <a_samp>

new a[2][2][2][2] = {
    {
        {
            {1, 2}, {3, 4}
        },
        {
            {5, 6}, {7, 8}
        }
    },
    {
        {
            {11, 12}, {13, 14}
        },
        {
            {15, 16}, {17, 18}
        }
    }
};

main() {
    for (new i = 0; i < sizeof a; i++) {
        for (new j = 0; j < sizeof a[]; j++) {
            for (new k = 0; k < sizeof a[][]; k++) {
                for (new l = 0; l < sizeof a[][][]; l++) {
                    printf("a[%d][%d][%d][%d] = %d", i, j, k, l, a[i][j][k][l]);
                }
            }
        }
    }
}
```

will produce:

```pawn
a[0][0][0][0] = 1
a[0][0][0][1] = 2
a[0][0][1][0] = 3
a[0][0][1][1] = 4
a[0][1][0][0] = 1 // should be 5
a[0][1][0][1] = 2 // should be 6
a[0][1][1][0] = 3 // should be 7
a[0][1][1][1] = 4 // should be 8
a[1][0][0][0] = 5 // should be 11
a[1][0][0][1] = 6 // should be 12
[...]

PR #57, opened 2015-04-13
Missing struct

Misiur:
Ref #43


Zeex:
Thanks, but I ended up dong my own fix for 4d as I was kind of curious how this indirection table stuff worked and eventually figured out what was wrong with them.

Issue #93, opened 2016-04-24
Random Seed


Crayder:
Currently, `IL_StandardRandom_seed` is set to the current time the first time `random` is called.

Instead, we could have a function to seed it ourselves. The function could simply set `IL_StandardRandom_seed`.

If we used the function before the first time we used `random` it would never use the time.
If we used it after, it would've used the time until we set it manually.
If we never used it, `random` would act as it normally does.

Crayder:
Could be as simple as this:

	static cell AMX_NATIVE_CALL core_srandom(AMX *amx,const cell *params)
	{
	    IL_StandardRandom_seed = (unsigned long)params[1];
	    return 1;
	}

With native line:

	{ "srandom",       core_srandom },

---------------------

I'll just make a pull request...

Crayder:
Of course however, since it's a native, only people who know that it exists here will be able to use it.

>native srandom(seed);

oscar-broman:
Are you using the runtime for something?

Crayder:
>Are you using the runtime for something?

Not sure what you mean...

I'm just suggesting adding a function to seed the random function.

oscar-broman:
Ok, but you do realize adding a function there will not make it work on a SA-MP server? The server's runtime can't be easily changed.

This just compiles an AMX, and the only information the AMX has about native functions are their names.

Crayder:
No actually I didn't think about that... xD
Not sure what I was getting at yesterday.

PR #94, opened 2016-04-24
Add 'srandom'

Crayder:
Addresses #93


Zeex:
As @oscarm-broman pointed out in #93, you can't change the AMX runtime built into the SA-MP server. This repo is basically for compiler patches only.

Issue #114, opened 2016-09-02
Pawn compiler 3.10.20160702 segfault on linux x64


xxmitsu:
Hello,
I'm using 
`Pawn compiler 3.10.20160702                     Copyright (c) 1997-2006, ITB CompuPhase`
version of your compiler, downloaded from the releases section.

with the following versions of glibc packages:

glibc-devel-2.17-106.el7_2.8.x86_64
glibc-devel-2.17-106.el7_2.8.i686
glib2-2.42.2-5.el7.x86_64
glibc-2.17-106.el7_2.8.x86_64
glib2-2.42.2-5.el7.i686
dbus-glib-0.100-7.el7.x86_64
glibc-headers-2.17-106.el7_2.8.x86_64
glibc-2.17-106.el7_2.8.i686
glibc-common-2.17-106.el7_2.8.x86_64

 however suddenly it started to segfault.

Full backtrace can be found here: http://pastebin.com/svN3Eznn

Thank you in advance for the help!

ziggi:
Linux version has more segfaults. I use little bit modified version without two lines (see pr #64). Without these lines compiler works pretty good (not perfect). You can download binary: [libpawnc.so](https://github.com/Open-GTO/Open-GTO/blob/master/compiler/libpawnc.so), [pawncc](https://github.com/Open-GTO/Open-GTO/blob/master/compiler/pawncc).

xxmitsu:
Thank you, it seems that is compiling fine with your modified version.
Do you happen to know what is the reason behind this ?

ziggi:
> Do you happen to know what is the reason behind this ?

See pr #115

xxmitsu:
Hello,

This issue can be closed when the specified fix will reach upstream.
Thanks again for investigating and finding a proper solution!

Zeex:
This was fixed by #115.

Issue #64, opened 2015-05-06
Odd crash on Linux and Windows XP (on all versions of the compiler)


ziggi:
Sometimes i got crash on 2078 line in sc3.c:
```C
        if (arg[argidx].numtags==1)     /* set the expected tag, if any */
          lval.cmptag=arg[argidx].tags[0]; // crash here
```
I can't show small Pawn code for reproduce this, because It is not always. Sometimes i change one symbol and get compiler crash, after it I change false on true in strcmp and fix it -_-.

On Windows 7 everything is always good.

Without these lines at first sight everything works good.

Zeex:
Can you show the output of `info locals` and `info args` commands in gdb?

ziggi:
```
Pawn compiler 3.10.20150503                     Copyright (c) 1997-2006, ITB CompuPhase


Program received signal SIGSEGV, Segmentation fault.
0x00007ffff799a7db in callfunction (sym=0x6eb400, lval_result=0x7fffffffdde0, 
    matchparanthesis=1) at /home/ziggi/dev/samp/development/pawn/source/compiler/sc3.c:2078
2078              lval.cmptag=arg[argidx].tags[0];
```
```
(gdb) info locals
nest_stkusage = 0
nesting = 1
locheap = 0
close = 0
lvalue = 32767
argpos = 0
argidx = 0
nargs = 0
heapalloc = 0
namedparams = 0
lval = {sym = 0x0, constval = 0, tag = 0, cmptag = 0, ident = 0 '\000', 
  boolresult = 0 '\000', arrayidx = 0x0}
arg = 0x6eaef0
arglist = "\002", '\000' <repeats 125 times>
arrayszlst = {next = 0x0, name = '\000' <repeats 31 times>, value = 0, index = 0}
taglst = {next = 0x0, name = '\000' <repeats 31 times>, value = 0, index = 0}
symret = 0x0
lexval = 32767
lexstr = 0x7ffff7bbc589 <pline+9> "_OnPlayerStateChange(playerid, newstate, oldstate);\r\n"
__PRETTY_FUNCTION__ = "callfunction"
```
```
(gdb) info args
sym = 0x6eb400
lval_result = 0x7fffffffdde0
matchparanthesis = 1
```

ziggi:
Maybe it will be useful too:
```
(gdb) p arg[argidx]
$16 = {
[...]

PR #115, opened 2016-09-05
Remove Linux crash lines (#64, #114)

ziggi:
Fix for #64, #114.

<del>Just removed two lines that give a crash. I've tested this very much and it pretty stable.</del>

<del>Why it stable? Because in latest Pawn these lines also does not exist: [sc3.c#L2331-L2333](https://github.com/compuphase/pawn/blob/master/compiler/sc3.c#L2331-L2333). And I think it's a good reason to remove this.</del>

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -2074,6 +2074,8 @@ static int nesting=0;
          */
       } else {
         arglist[argpos]=ARG_DONE; /* flag argument as "present" */
+        if (arg[argidx].ident!=0 && arg[argidx].numtags==1)
+          lval.cmptag=arg[argidx].tags[0];  /* set the expected tag, if any */
         lvalue=hier14(&lval);
         assert(sc_status==statFIRST || arg[argidx].ident== 0 || arg[argidx].tags!=NULL);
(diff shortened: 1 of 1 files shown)


oscar-broman:
Doesn't this remove tag checking for forwarded functions, or functions used before they are declared?

Y-Less:
The latest PAWN compiler is almost a completely different language.  I wouldn't trust that as an example of what can be changed.  There's also no guarantee that they haven't just been moved or implemented differently.  I'm with Oscar - do you know what they were there for, removing lines does imply removing functionality and they clearly do something with tags.

oscar-broman:
The problem is not that line. The problem is that `arg[argidx].tags` has an invalid value. When I inspected it in gdb, I saw `0x4` - and it's a pointer!

Something has gone wrong before it reached this line.

oscar-broman:
I changed the above if-statement to this: https://github.com/compuphase/pawn/blob/6d82fa4bfa3df019f8144dcb75b994240f8b9a7e/compiler/sc3.c#L2236

It fixes the segfault.

ziggi:
Seems to it really works, thanks!

Issue #125, opened 2016-12-15
Negative numbers in emit constants


Y-Less:
```pawn
#emit CONST.pri -5
```

Doesn't work.  The only way to do that is with the HEX equivalent:

```pawn
#emit CONST.pri 0xFFFFFFFB
```

YashasSamaga:
#127

YashasSamaga:
#128

YashasSamaga:
#128

PR #128, opened 2016-12-26
allow negative numbers in #emit & changes to public variable behaviour

YashasSamaga:
allows negative numbers in #emit

fix for #125

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -4690,7 +4690,7 @@ static int testsymbols(symbol *root,int level,int testlabs,int testconst)
       /* a variable */
       if (sym->parent!=NULL)
         break;                      /* hierarchical data type */
-      if ((sym->usage & (uWRITTEN | uREAD | uSTOCK))==0) {
+      if ((sym->usage & (uWRITTEN | uREAD | uSTOCK | uPUBLIC))==0) {
         if (testconst)
           errorset(sSETPOS,sym->lnumber);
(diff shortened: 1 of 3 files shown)


VVWVV:
Yes, you are allow to use negative numbers in #emit, but this is only for integer numbers, not floating.
For example, I did this: https://github.com/VVWVV/pawn/commit/880fd101fee5f422e35da3233f5e80f30dc539c2

YashasSamaga:
I don't think someone will ever need to pass a rational value to an assembly instruction.

Y-Less:
I guess they would if they wanted to do:

```pawn
#emit PUSH.C 0.5
#emit PUSH.C 7.7
#emit PUSH.C 8
#emit SYSREQ.C floatmul
```

But that seems more likely to cause confusion and mistakes since there is no tag checking so a stray `.` won't be flagged and will vastly alter the result.

Zeex:
@YashasSamaga Mind fixing Appveyor build and indentation?

YashasSamaga:
1. AppVeyor requires the code to abide by the C89 declaration rules. Moved the s2 declaration to the start of the block.

2. Indentation Fixed. I have a doubt with the indentation for line 755 in sc6 ('else'). I am not sure how you prefer to indent else with single statement. You can edit it.

Zeex:
Great, thanks. 

Yes, Visual Studio doesn't support C99, and in fact the whole codebase is written in C89, so I try to stick to it too (I think it was originally done for maximum portability).

PR #135, opened 2017-01-22
Fix warning 204/203 in sizeof and tagof operators

VVWVV:
````pawn
main() {
	new Float:tmp;
	printf("tagid: %08x", tagof tmp);

	new array[12];
	printf("sizeof: %d", sizeof array);
}
````

The compiler gives the warnings:
```
warning 204: symbol is assigned a value that is never used: "array"
warning 203: symbol is never used: "tmp"
```

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -1370,6 +1370,7 @@ static int hier2(value *lval)
         error(224,st);          /* indeterminate array size in "sizeof" expression */
     } /* if */
     ldconst(lval->constval,sPRI);
+    markusage(sym,uREAD);
     while (paranthese--)
       needtoken(')');
     return FALSE;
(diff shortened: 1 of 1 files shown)


Y-Less:
I'm not seeing the problem - they aren't used.  `sizeof` is a compile-time operator that compiles to a constant, so the compiler is correct - they are not used anywhere in final code.  If you create a 200 element array, then never access that array, you have by definition not used the array.  Getting the size or the tag isn't using it, just retrieving meta data.  The first warning is absolutely correct because `symbol is assigned a value that is never used` is true - you don't use the value.

Y-Less:
Point being, the warning in this case will catch far more real mistakes in code than it will trigger false positives from very unusual code.

VVWVV:
You're absolutely right.

Issue #134, opened 2017-01-22
Proposal: enum limited to single file scope


Yiin-:
So it could be possible to do

a.pwn
```
static enum Foo {
  E_FOO,
  E_BAR
}
```

b.pwn
```
static enum Foo {
  E_FOO = 2,
  E_BAR
}
```

main.pwn
```
#include "a.pwn"
#include "b.pwn"

enum Foo {
  E_FOO,
  E_BAR
}
```

Zeex:
This will be available in the next release. It's mostly done but needs some polish.

You can try now: https://ci.appveyor.com/project/Zeex/pawn/build/149/artifacts

PR #136, opened 2017-01-23
#134: Proposal: enum limited to single file scope

VVWVV:
See #134

Example:
```pawn
enum static {
     s_1,
     s_2
};
```
or
```pawn
enum static Abc {

};
static var1[Abc];
```
or
```pawn
static enum Abc {

};
static var1[Abc];
```

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -89,7 +89,7 @@ static void declglb(char *firstname,int firsttag,int fpublic,int fstatic,
                     int stock,int fconst);
 static int declloc(int fstatic);
 static void decl_const(int table);
-static void decl_enum(int table);
+static void decl_enum(int table,int enum_static);
 static cell needsub(int *tag,constvalue **enumroot);
 static void initials(int ident,int tag,cell *size,int dim[],int numdim,
                      constvalue *enumroot);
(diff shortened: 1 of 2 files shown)


Y-Less:
Out of interest, what happens if you do:

f1.pwn:

```pawn
static enum X
{
    Y,
    Z,
    A
};

new gVar[X];
```

f2.pwn:

```pawn
#include "f1.pwn"

main()
{
    // This:
    printf("%d", gVar[A]);
    // Or:
    printf("%d", gVar[0]);
    // Or:
    printf("%d", gVar[X:0]);
}
```

?  Basically, a static enum on a global variable.

VVWVV:
@Y-Less, this is the wrong code! You need to use `enum static`, not `static enum`.

And when i compile, the compiler gives:
```shell
error 017: undefined symbol "A"
warning 213: tag mismatch
```

Y-Less:
Why?  `static stock const`, `const static stock`, `const stock static` etc - all valid.  Why is keyword order important only here?  I have no doubt I won't be the only person to make that mistake - to me `static enum` flows much better because it describes what it is in English - it is a static enumeration.

Good that those examples don't work though.

VVWVV:
Ok! I'll do it. Thanks!

karimcambridge:
Yea, you don't do "class static namehere", you do "static class namehere".

It's a serious convention.

VVWVV:
I did it! Now you can use both: `static enum` and/or `enum static`.

VVWVV:
> Out of interest, what happens if you do:
> 
> f1.pwn:
> ```pawn
> static enum X
> {
>     Y,
>     Z,
>     A
> };
> 
> new gVar[X];
> ```
> f2.pwn:
> ```pawn
> #include "f1.pwn"
> 
> main()
> {
>     // This:
>     printf("%d", gVar[A]);
>     // Or:
>     printf("%d", gVar[0]);
>     // Or:
>     printf("%d", gVar[X:0]);
> }
> ```
> ? Basically, a static enum on a global variable.

Yes, we have access to global variable, but we don't have access to the elements in static enumerations. So if we use the elements of the static enumerations, the compiler gives the error:
```shell
error 017: undefined symbol "A"
```
code:
````pawn
printf("%d", gVar[A]);
````

**Comment by Y-Less** [comment]
You already said that, but thanks.

**Comment by Zeex** [comment]
Thanks! Merged this with some small adjustments (mostly cosmetic).

However, the error reporting needs some improvement. I could not trigger that new error 92 by declaring a local static enum. Instead, the compiler gave me another (less descriptive) error:

```pawn
main() {
	static enum e {}; // error 001: expected token: "-identifier-", but found "enum
	enum static e {}; // error 001: expected token: "{", but found "static"
}
```

@VVWVV Are you willing to make another PR improving this aspect?

VVWVV:
@Zeex, ok! I will fix it.

VVWVV:
I've fixed it. But I think, we can declare the static enums in blocks without errors, because the static operator can declare a variable in a block and in the global scope.

Zeex:
I guess it would make sense if `static enum` worked everywhere where `enum` works.

PR #139, opened 2017-01-27
Fix a compiler crash with emit

VVWVV:
If we don't use functions with #emit, the compiler will crash. But it doesn't work with public functions.

Example:
````pawn

stock abc()
{
#emit PUSH.c 6
#emit PUSH.c 4
#emit SYSREQ.c random
#emit STACK 8
#emit RETN
return;
}

main()
{

}
````

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1318,6 +1318,8 @@ static int command(void)
             error(17,str);        /* undefined symbol */
           } else {
             if (sym->ident==iFUNCTN || sym->ident==iREFFUNC) {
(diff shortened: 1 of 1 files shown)


Zeex:
That piece of code worked fine for me. Can you show an example where it crashes?

VVWVV:
You have already fixed this crash (commit f9e9c181dd49aa8c3f7c0b7292dd2dce9f9f4bcf)

Zeex:
Oh, didn't realize that. Thanks anyway!

PR #143, opened 2017-02-01
Minor fixes/adjustments for the build script

Daniel-Cortez:
Just as the title says, the bugs fixed in this PR are minor and only show up if script `compiler/CMakeLists.txt` is used from another CMake script (with `add_subdirectory`).

1. When generating a project CMake can't find file `version.h`. `CMAKE_CURRENT_BINARY_DIR` and `CMAKE_BINARY_DIR` variables are used as if they are interchangeable while they actually aren't.
   `configure_file` command saves files into `CMAKE_CURRENT_BINARY_DIR` directory by default, not into `CMAKE_BINARY_DIR`.
2. For some reason when I compile the project (with `compiler/CMakeLists.txt` used from another CMake script too) with MinGW, it builds libpawnc.dll instead of pawnc.dll and pawncc.exe won't work because it needs pawnc.dll, without the "lib" prefix, so I removed it in the script.
   The prefix should probably be removed when building with Watcom C and/or Borland C as well, I just haven't tried building with these compilers yet.

### DIFF
--- a/source/compiler/CMakeLists.txt
+++ b/source/compiler/CMakeLists.txt
@@ -50,7 +50,8 @@ if(UNIX)
   link_libraries(pthread)
 endif()
 
-configure_file(version.h.in version.h)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/version.h.in
+               ${CMAKE_CURRENT_BINARY_DIR}/version.h)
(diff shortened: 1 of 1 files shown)


Zeex:
Thanks.

Issue #78, opened 2015-11-17
warning 226: a variable is assigned to itself .. bug


karimcambridge:
```
enum e_System
{
    bool:e_Can_Team_DM,
    bool:e_Can_Ticket,
    bool:e_Can_Arrest,
    bool:e_Can_pRobbery,
    bool:e_Can_pRape,
    bool:e_Can_pTaze,
    bool:e_Can_pKidnap,
    bool:e_Can_Use_Weapons,
    bool:e_Can_pMedics,
    bool:e_Can_Stickup,
    bool:e_NameTagStatus,
    bool:e_CanPlayAnimations
};
new Event_System[e_System];
```

```
Event_System[e_Movement] = !Event_System[e_Movement];
Event_System[e_Can_Team_DM] = !Event_System[e_Can_Team_DM];
```

etc.. all generate this error

```
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24328) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24349) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24360) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24371) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24382) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24393) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24404) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24415) : warning 226: a variable is assigned to itself (symbol "Event_System")
C:\Users\Documents\SAMP\SAMP Server\gamemodes\gm.pwn(24426) : warning 226: a variable is assigned to itself (symbol "Event_System")
[...]

PR #150, opened 2017-02-12
fixes false warning (226) triggers for arrays

YashasSamaga:
Fixes Issue #78

Creates a member variable called 'value_modified' in the value struct
which keeps track if the symbol was modified.

clear_value function sets the member variable to FALSE by default.

When a !,~ or - operation is performed on the symbol, the member is set
to TRUE.

The self-assignment warning is not triggered for arrays unless the
value_modified variable is not set.

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -242,6 +242,7 @@ typedef struct s_value {
   int cmptag;           /* for searching symbols: choose the one with the matching tag */
   char ident;           /* iCONSTEXPR, iVARIABLE, iARRAY, iARRAYCELL,
                          * iEXPRESSION or iREFERENCE */
+  char value_modified;	/* set to TRUE when value is negated/inverted */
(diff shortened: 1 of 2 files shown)


YashasSamaga:
This introduces a new variable 'value_modified' in the value structure just for the sake of fixing the bug. So I doubt if this would be a good fix.

In the future, the same member can be used for other purposes. Maybe rename the member and use it as a bitmask?

Zeex:
Thanks for the PR! I fixed this in a slightly different way in https://github.com/Zeex/pawn/commit/43cd36b891a674df928eaf587469e85dfa36bbdb which I think is a little bit better as it doesn't add a new member variable to the struct.

Dayrion:
I've the same bug and the last version of Zeex compiler:
`f_GeneralValues[gvPoliceStation] = !f_GeneralValues[gvPoliceStation];
bool:gvPoliceStation`

Issue #129, opened 2017-01-03
ternary operator bug when using strings


irinel1996:
I don't know why but the next code doesn't work as it should:

```
new gender[10];

gender = (p_Gender[playerid] = 0 ? "Male" : "Female");
```
If `p_Gender[playerid] != 0` I get "Femal" instead of "Female", there are enough cells to store it. I don't know what's going on...

Thank you.

karimcambridge:
That isn't good code at all.

Also to use strings in ternary, wrap it around () brackets.

```
new gender[10];
format(gender, 10, "%s", p_Gender[playerid] == 0 ? ("Male") : ("Female"));
```

YashasSamaga:
Why is that code not good? I use it everywhere (with brackets of course).

Jeroen52:
Not sure about the context on how you'd use the `gender` string, but from what I can see I just recommend you to put it in-line when formatting a string.

I'd do it something like this.
`format(mysql_mediumString, sizeof(mysql_mediumString), ""COL_WHITE"%s\nSet House Password\nHouse Keys\nManage players inside house\n \n"COL_RED"<< Back", ((hInfo[pInfo[playerid][pInHouse]][houseLocked]) ? ("Unlock house") : ("Lock House")));`

Jeroen52:
One more tip.

You could also make an array and enum for player info, so you can change `p_Gender[playerid]` to `pInfo[playerid][gender]`, or `pInfo[playerid][cash]`, `pInfo[playerid][score]`.

Y-Less:
Because you didn't do `!=`, you just did `=`, which is very different.

irinel1996:
@Y-Less just a typo, in the gamemode the condition is written correctly.

@Jeroen52 @YashasSamaga brackets are necessary with the default compiler, they aren't with ZeeX's compiler as far as I know. See https://github.com/Zeex/pawn/commit/764b373e38a33ccd18b7fd50fc02539f2d60d8b1

@Jeroen52 I know, I wrote that example to make the code more readable.

Y-Less:
Try:

```pawn
gender = p_Gender[playerid] ? "Female" : "Male";
```

If that works, I have a hunch as to the problem - it is only copying the length of the first string found.  Of course, this could also crash if that's the case; or it could still not work.

irinel1996:
@Y-Less it works now. I guess you're right about the copied length, the same thing happens with other strings.

Y-Less:
[...]

PR #151, opened 2017-02-12
triggers an error for issue #129

YashasSamaga:
cond ? array1 : array2

Triggers an array if the size of array1 and array2 do not match.


YashasSamaga:
This does not really fix the issue but it adds code to trigger an error instead of allowing the buggy code to compile successfully.

Zeex:
I had to revert this because of #198, feel free to submit a new PR with fixes for that issue.

Issue #117, opened 2016-09-20
Multi-Dimensional initialisation bug


Dayvison:
With this works.

``` pawn
#define DEF 3
new Float:test[DEF][3][10] = {
    {
        { 0.00, 1.01, 2.02, 3.03, 4.04, 5.05, 6.06, 7.07, 8.08, 9.09 },
        { 0.010, 1.011, 2.012, 3.013, 4.014, 5.015, 6.016, 7.017, 8.018, 9.019 },
        { 0.020, 1.021, 2.022, 3.023, 4.024, 5.025, 6.026, 7.027, 8.028, 9.029 }
    },
    {
        { 0.030, 1.031, 2.032, 3.033, 4.034, 5.035, 6.036, 7.037, 8.038, 9.039 },
        { 0.040, 1.041, 2.042, 3.043, 4.044, 5.045, 6.046, 7.047, 8.048, 9.049 },
        { 0.050, 1.051, 2.052, 3.053, 4.054, 5.055, 6.056, 7.057, 8.058, 9.059 }
    },
    {
        { 0.060, 1.061, 2.062, 3.063, 4.064, 5.065, 6.066, 7.067, 8.068, 9.069 },
        { 0.070, 1.071, 2.072, 3.073, 4.074, 5.075, 6.076, 7.077, 8.078, 9.079 },
        { 0.080, 1.081, 2.082, 3.083, 4.084, 5.085, 6.086, 7.087, 8.088, 9.089 }
    }
};
```

Output:

``` pawn
new test[3][3][10] = {
        {
                { 0.000000, 1.009999, 2.019999, 3.029999, 4.039999, 5.050000, 6.059999, 7.070000, 8.079999, 9.090000 }
                { 0.009999, 1.011000, 2.012000, 3.013000, 4.013999, 5.014999, 6.015999, 7.017000, 8.017999, 9.019000 }
                { 0.019999, 1.021000, 2.022000, 3.023000, 4.024000, 5.025000, 6.026000, 7.026999, 8.027999, 9.029000 }
        }
        {
                { 0.029999, 1.031000, 2.032000, 3.032999, 4.033999, 5.034999, 6.035999, 7.037000, 8.038000, 9.038999 }
                { 0.039999, 1.041000, 2.042000, 3.042999, 4.044000, 5.045000, 6.046000, 7.046999, 8.048000, 9.048999 }
[...]

PR #152, opened 2017-02-13
fixes multi-dimesional array initilization bug

YashasSamaga:
Solves the bug described in issue #117

new [][2][3] = { //default values };
The compiler does not create enough room for the indirection tables and
hence adjust_indirectiontables overwrites the default values leaving the
array wrongly initilized.

Adds a generic code which works for any n-dimensional arrays (n = 1,
2,3,4,5... infinity). The code fixes the issue for any dimension but the
number of dimensions is limited to 4 (sDIMEN_MAX). A review of this code
won't be required if the limit is increased in the future.


YashasSamaga:
Not sure if this is a good fix. This fixes but allows the bad calc_array code to remain in the initials.

PR #153, opened 2017-02-13
fixes memory corruption with multi-dimensional arrays

YashasSamaga:
static const Test[2][][2] = { { { 9, 8 } },  { { 6, 7 } } };

The above code causes a memory corruption as the compiler does not
correctly calculate the size of the array.

This commit adds code to fix it.


YashasSamaga:
I was thinking of removing that code from initials and moving it to calc_arraysize instead. calc_arraysize is supposed to return 0 when it is not able to calculate the full size of the array but it doesn't when the dimensions are more than 2.

Zeex:
Thanks!

PR #174, opened 2017-06-08
Add 'Installation on openSUSE/SLES' section to README

mschnitzer:
I'd like to suggest to add an `Installation on openSUSE/SLES` section to the README. One year ago, I started packaging your project for openSUSE distributions. The plan was to provide a package that everyone can install with openSUSE's package manager `zypper`. The package gets updated to the newest version when you release a new update. Users can then obtain the newest version when they update their distribution (command: `zypper up` ; would be equal to `apt-get update` on Debian/Ubuntu)

This makes it easier for users that have more than one computer/server to update to the newest version of this project.

I also planned to submit this package to the openSUSE distribution so that new released openSUSE distributions (next will be `Leap 42.3`) provide that package by default.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -13,6 +13,18 @@ What is this?
 This is a modified copy of the Pawn compiler version 3.2.3664 by Compuphase that
 fixes some bugs and adds a few features.
 
+Installation on openSUSE/SLES
+---------------------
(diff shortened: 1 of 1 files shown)


Zeex:
Thanks.

PR #180, opened 2017-08-19
emit/__emit operator

VVWVV:
This PR implements the `emit/__emit` operator, a modern alternative for the `#emit` directive.


Daniel-Cortez:
### Pros and cons of the `asm` operator over the `#emit` directive
:heavy_plus_sign: `asm` is an operator, not a preprocessor directive, so it can be used in expressions and marcos.<br/>
:heavy_plus_sign: Argument types are checked by the compiler (e.g. branch instructions like `JUMP`, `JZER`, `JNZ` etc. only take a label name as an argument, otherwise the compiler would issue an error).

:heavy_minus_sign: Not a part of Pawn 3.2 standard.
<br/>
<br/>

### Syntax
The `asm` operator has 2 syntaxes: single-line and miltiline.

#### Single-line:
```C++
asm const.pri 1
```
This syntax is very simple. Only one `asm` statement per line and one AMX instruction per statement are allowed.
To output more than one instruction, either use multiline syntax (see below) or multiple single-line `asm` statements:
```C++
// Swap `a` and `b`
new a = 4, b = 5;
asm load.s.pri a
asm load.s.alt b
asm stor.s.pri b
asm stor.s.alt a
```

#### Multiline:
```C++
asm { const.pri 1 }
```
Multiple instructions in one multiline `asm` statement are allowed, but with one instruction per line only.
```C++
new a = 4, b = 5;
asm {
    load.s.pri a
    load.s.alt b
    stor.s.pri b
    stor.s.alt a
}
```
To put multiple AMX instructions in one line, put each of them into separate multiline `asm` statements, one instruction per statement:
```C++
new a = 4, b = 5;
asm { load.s.both a b } asm { stor.s.pri b } asm { stor.s.alt a }
```
Also the multiline `asm` operator can be used to make expressions. `asm` expression returns the value contained in the `PRI` register.
```C++
new a = asm { const.pri 1 }; // a = 1
```
Another example:
```C++
printf(
    "Free stack space: %d",
    (asm{lctrl 4}, asm{heap 0}, asm{sub})
);
```
This can be useful for function-like macros:
```C++
#define GetFreeStackSpace() (                       \
    asm{lctrl 4},                                   \
    asm{heap 0},                                    \
    asm{sub}                                        \
)

main()
{
    printf("Free stack/heap space: %d", GetFreeStackSpace()); // The return value is used in `printf()`
    GetFreeStackSpace();
}
```

Y-Less:
This is awesome, but why not stick to `emit` naming?  IMHO it would be much clearer that `#emit` `@emit` (in amx_assembly) and `emit` from this are all related.  There is also already `__asm` in amx_assembly that actually outputs x86 assembly, and `asm` is more likely to be in use in existing codebases than `emit`.  It would also reduce the number of unique keywords.

Daniel-Cortez:
> This is awesome, but why not stick to emit naming? IMHO it would be much clearer that #emit @emit (in amx_assembly) and emit from this are all related.

Well, that's debatable. In my opinion, `emit` may cause more confusion with `#emit` and `@emit` rather than help by making them all related. I'm not a veteran of the SA-MP community like you, though, so you probably know better.

Y-Less:
Or `__emit`, since double-underscore prefixes are meant to be reserved for compiler things only, there's no chance of conflicts with conforming code.

VVWVV:
@Y-Less, yes, you are absolutely right. Ok, I'll change the name.

Daniel-Cortez:
The name `asm` was influenced by the `asm` operator from C/C++ (well, obviously). Initially the operator was called `_asm` in order to avoid conflicts, but then we (and by "us" I mean @VVWVV and myself; I was reviewing @VVWVV's code from time to time before making it into this PR) decided to stick with `asm` for the following reasons:
1. It's simpler to type (unlike `@asm`, `__asm`, `__emit` etc.)
2. It's an improvement for the language itself, not for a third-party library where names starting with clunky prefixes like "__" or "@" could be (more or less) acceptable.

Anyway, avoiding conflicts seems reasonable as well, so maybe keeping `asm` with `__asm` as a fallback would be the best option? Or "emit" and "__emit" if both of the former conflict with something else? That way we will both provide compatibility and not sacrifice the operator's name for third-party code completely.

Also, not sure about that, but maybe we should start a poll or something to determine the best option?

oscar-broman:
It makes more sense to use `emit` instead of `asm`, since the Pawn creators decided to go with `#emit` and not `#asm` (or something else).

This PR adds an operator closely related to `#emit`, so why not name it `emit`?

Names related to pre-processor directives:
`#if` - `if`
`#else` - `else`
`#assert` - `assert`
`#define` - `defined`
`#warning` - `#pragma warning`
`#emit` - `emit` (?)

Daniel-Cortez:
Ok then, I guess I was wrong about confusion between `__emit`, `#emit` and `@emit`; consistency with directive names makes much more sense.
Also both `asm` and `__asm` are already used as macro names in amx_assembly, so, IMHO, using `emit` and `__emit` is obviously the best option.

Y-Less:
@Zeex could you have a look at this (and the other pull requests) please?

Zeex:
This seems like a useful thing, thanks for the PR. Nicely done. 

I too think that `emit` / `__emit` is a better name. I was going to make a poll at some point because disagreements in this thread but  Y_Less convinced everybody pretty fast 😃

Issue #184, opened 2017-09-10
Documentation for state transitions appears OOB


Y-Less:
Example (compiled with `-r`):

```pawn
#include <a_samp>

/**

**/

FuncWithState() <Automata : StateA>
{
}

/**

**/

FuncWithState() <Automata : StateB>
{
}

/**

**/

main()
{
	state Automata : StateB;
	FuncWithState();
}
```

Produces (amongst other data):

```xml
<?xml version="1.0" encoding="ISO-8859-1"?>
<?xml-stylesheet href="file:///D|/work/SA-MP/pawno/xml/pawndoc.xsl" type="text/xsl"?>
<doc source="D:\work\SA-MP\gamemodes\statedoc.pwn">
	<assembly>
		<name>statedoc.pwn</name>
	</assembly>
	<members>
		<member name="M:main" syntax="main()">
			<attribute name="entry"/>
			<stacksize value="3"/>
			<dependency name="FuncWithState"/>
			<transition target="pQ"/>

		</member>

	</members>
</doc>
```

Which includes:

```xml
			<transition target="pQ"/>
```

Pretty sure that should be:

```xml
			<transition target="StateB"/>
```

I've had worse garbarge that included quotes, screwing up the XML, and what looked like other old tags.

PR #185, opened 2017-09-10
Document correct state transition target name.

Y-Less:
TBH, I'm not fully sure how to set up a local build on Windows, so I didn't bother.  I'm 99% sure the code is correct, so I'm just relying on your CI to confirm this.


Y-Less:
Fix for #184

Issue #159, opened 2017-03-15
ignored constant sym param


Dayvison:
Code:
```c++
#if defined a
	#error "a"
#else
	#error "b"
#endif
main() { }
```
Command used:
`pawncc.exe m.pwn a=1`

Zeex compiler:
```
Pawn compiler 3.10.2                    Copyright (c) 1997-2006, ITB CompuPhase

m.pwn(4) : fatal error 111: user error: "b"


Compilation aborted.


1 Error.
```
Original compiler:
```
m.pwn(2) : fatal error 111: user error: "a"


Compilation aborted.

Pawn compiler 3.2.3664                  Copyright (c) 1997-2017, ITB CompuPhase


1 Error.
```

YashasSamaga:
The compiler ignores constants which are passed as arguments. 

1. The compiler parses the compiler options at the very begining.
2. The compiler deletes the entire global symbol table before parsing.

So any constants which were added when the compiler options was being parsed is lost.

The bug is in the following line: https://github.com/Zeex/pawn/blob/master/source/compiler/sc1.c#L583

Southclaws:
+1 I use this and I only found about this bug today, it would really screw up my build process if I migrated to this compiler (which I wanted to for SS).

@YashasSamaga could you submit a pull request with the proposed fix?

PR #189, opened 2017-09-29
fixes deletion of constants passed a compiler args during parse

YashasSamaga:
fix for issue #159 

When constants are passed as compiler arguments (`pawnc test.pwn myvar=10`), the compiler adds them to the symbol table using [add_constant](https://github.com/Zeex/pawn/blob/master/source/compiler/sc1.c#L1203). Before the compiler begins to parse the script, it[ removes all the symbols from the symbol table except the builtin/predefined constants](https://github.com/Zeex/pawn/blob/master/source/compiler/sc1.c#L583). While most of the symbols are added back to the symbol table as the script is parsed, the symbols passed as compiler arguments won't be added again (as the compiler options are not parsed again).

This PR changes the `add_constant` function call with `add_builtin_constant` so that the constants passed through the argument list is treated as a predefined constant (and hence not deleted).

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -1200,7 +1200,7 @@ static void parseoptions(int argc,char **argv,char *oname,char *ename,char *pnam
       } /* if */
       strlcpy(str,argv[arg],i+1);       /* str holds symbol name */
       i=atoi(ptr+1);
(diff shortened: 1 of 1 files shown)


Zeex:
Thanks!

Issue #187, opened 2017-09-28
deprecation messages leave extra lines in compiler output


Y-Less:
I have this:

```pawn
#pragma deprecated Use formatex instead
forward va_formatex(output[], size = sizeof(output), const fmat[], va_:STATIC_ARGS);
```

Which gives the following output when the function is used:

```
D:\work\SA-MP\gamemodes\deprecated.pwn(42) : warning 234: function is deprecated (symbol "va_formatex") Use formatex instead

D:\work\SA-MP\gamemodes\deprecated.pwn(53) : warning 234: function is deprecated (symbol "va_formatex") Use formatex instead

```

There are blank lines placed between the error lines.  I've tried a few ways to get rid of them - changing line endings and escaping them, but neither work.  With no message, the spacing is correct:

```pawn
#pragma deprecated
forward va_formatex(output[], size = sizeof(output), const fmat[], va_:STATIC_ARGS);
```

Which gives the following output when the function is used:

```
D:\work\SA-MP\gamemodes\deprecated.pwn(42) : warning 234: function is deprecated (symbol "va_formatex")
D:\work\SA-MP\gamemodes\deprecated.pwn(53) : warning 234: function is deprecated (symbol "va_formatex")
```

Dayvison:
same for  `#warning`

YashasSamaga:
After reading the warning/pragma deprecated directive, the compiler skips white spaces (or until it encounters a null character). It passes the remaining line string as it is to the error function. This line string would include a new line character which was added by the scripter.

However, the compiler adds also adds a new line by itself at the end of every warning and pragma deprecated. Therefore, the compiler would print two new lines.

Zeex:
This was fixed by #192 and https://github.com/Zeex/pawn/commit/e7ecf386f8e82ed03a3627712d2064940c61aeb4.

PR #192, opened 2017-09-30
remove extra new lines from #warning and #pragma deprecated

YashasSamaga:
The `#warning` and `#pragma deprecated` directives print the new line characters which are part of the user entered string. As the message formats for both these already contain a new line character at the end, the aforementioned directives printed two new lines.

This commit adds code to discard the new line characters from user string.

In case of an allocation failure, the compiler falls back to the old code.

Resolves Issue: #187

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1494,10 +1494,14 @@ static int command(void)
       lptr++;
     if (!SKIPPING) {
       char *usermsg=(char*)malloc(strlen(lptr)+1);
-      strcpy(usermsg,lptr);
-      usermsg[strcspn(usermsg,"\r\n")]='\0';
-      error(237,usermsg);  /* user warning */
-      free(usermsg);
+      if(usermsg!=NULL) {
+        strcpy(usermsg,lptr);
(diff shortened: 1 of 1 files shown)


Y-Less:
That's only going to work on Windows, you need to check for all trailing new line types.

YashasSamaga:
The code replaces the first occurrence of '\r' or '\n' with null character. For Windows based systems, the '\r' will be replaced by '\0' (we don't care about the '\n' after the null character) and for Linux based systems, the '\n' will be replaced by '\0'. Modern Mac computers use '\n' for terminating lines and older versions of Mac use '\r'.

The code in the PR replaces the first '\r' or '\n' (whichever comes first) with '\0'.

Y-Less:
OK, maybe I should have read its documentation first, instead of assuming it was like `strpos`.  Sorry.  How does this behave:

```pawn
#warning my \
warning \
message
```

?

YashasSamaga:
The `readline` function concatenates the three lines and passes the "my warning message" as a whole line for parsing, i.e: the `lptr` will point to the concatenated line in the memory. It does not matter to any parsing code whether multiple lines were used.

PR #199, opened 2017-10-22
Performance improvements

maddinat0r:
This PR improves overall compile speed by a factor of ~10.
Special thanks to @Daniel-Cortez for several changes.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,10 +1,11 @@
 version: '{build}'
+image: Visual Studio 2015
 
 configuration:
   - RelWithDebInfo
 
 before_build:
-  - cmake -G "Visual Studio 10 2010" source/compiler -DCPACK_GENERATOR=ZIP
+  - cmake -G "Visual Studio 14 2015" source/compiler -DCPACK_GENERATOR=ZIP
 
 build_script:
   - cmake --build . --config %CONFIGURATION%
--- a/source/compiler/CMakeLists.txt
+++ b/source/compiler/CMakeLists.txt
@@ -60,6 +60,8 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 # The Pawn compiler shared library
 set(PAWNC_SRCS
+  hashmap/hashmap.c
+  hashmap/hashmap.h
   libpawnc.c
   lstring.c
   lstring.h
--- a/source/compiler/hashmap/LICENSE
+++ b/source/compiler/hashmap/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 David Leeds
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/source/compiler/hashmap/hashmap.c
+++ b/source/compiler/hashmap/hashmap.c
@@ -1,0 +1,689 @@
+/*
+ * Copyright (c) 2016-2017 David Leeds <davidesleeds@gmail.com>
+ *
+ * Hashmap is free software; you can redistribute it and/or modify
+ * it under the terms of the MIT license. See LICENSE for details.
+ */
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <string.h>
+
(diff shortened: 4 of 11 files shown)


Daniel-Cortez:
The [initial size of the hashmap](https://github.com/maddinat0r/pawn/blob/7363feb6126270cc413fae434a97076f909a15d8/source/compiler/sc1.c#L931) still probably has to be changed.
IMHO, even if the hashmap growth process (memory reallocation, rehashing) takes time, there's no real need to pre-allocate 8 billion slots for every compiled script. The biggest SA-MP gamemode script I tried to compile with this made the compiler use roughly 6k slots at most.

Southclaws:
This build appears to break argument-defined constants:

```pawn
#if defined SOME_CONST
#error defined
#else
#error not defined
#endif
```

original:
```bash
$ pawncc.exe test.pwn SOME_CONST=
Pawn compiler 3.10.2                    Copyright (c) 1997-2006, ITB CompuPhase

test.pwn(2) : fatal error 111: user error: defined
```
This PR:

```bash
$ pawncc.exe test.pwn SOME_CONST=
Pawn compiler 3.10.2                    Copyright (c) 1997-2006, ITB CompuPhase

test.pwn(4) : fatal error 111: user error: not defined
```

maddinat0r:
I can't reproduce this with this PR rebased against 3.10.3. There probably was a bug before 3.10.3 somewhere related to your issue.

Zeex:
That is a huge improvement in performance! But I think it needs more testing before getting merged as it potentially could break a lot of things. 

Unfortunately I can't test it as I don't do any scripting at big enough scale besides small test scripts, but I can create a separate release for this version and have other people download and test it.

appveyor.yml
```diff
@@ -1,10 +1,11 @@
 version: '{build}'
+image: Visual Studio 2015
 
 configuration:
   - RelWithDebInfo
 
 before_build:
-  - cmake -G "Visual Studio 10 2010" source/compiler -DCPACK_GENERATOR=ZIP
+  - cmake -G "Visual Studio 14 2015" source/compiler -DCPACK_GENERATOR=ZIP
```
Zeex: Any reason for updating to 2015? Visual Studio 2010 runtime was more widespread when this file was originally created so it was used to avoid troubles with missing MSVC runtime DLLs.

Zeex:
I created a new download here: https://github.com/Zeex/pawn/releases/tag/v3.10.3-performance

appveyor.yml
maddinat0r: The AppVeyor build failed because the new hashmap depends on `stdbool.h`, and VS 2010 didn't ship it. But maybe VS 2012 also has it, I didn't look into that.

Y-Less:
I ran the full YSI test suite (on my broken mid-edit local copy) and didn't have any new issues/fails, so that's reassuring to me at least.

spacemud:
Reporting in: my fairly large mode (probably similar to SS in scale) compiles and runs fine with this fork.

Y-Less:
I was intending to test SS as well, but maybe @Southclaws can weigh in?

Dayvison:
Compiled well here, amx have no change
36 seconds to 2.6 xD

Jeroen52:
I have upgraded my continous intergration with zeex:performance (a.k.a. v3.10.3-performance).

This is the Dockerfile I'm using.
```Dockerfile
FROM centos:7

MAINTAINER jeroen@jeroendeneef.com

ENV VERSION 3.10.3-performance
ENV PAWNCFOLDER ~/pawn

LABEL	name="Zeex's pawn compiler" \
				vendor="Zeex" \
				license="zLib/libpng" \
				build-date="20171025"

RUN yum -y install \
				deltarpm epel-release; \
		yum -y groupinstall \
				'Development Tools'; \
		yum -y install \
				kernel-devel git glibc.i686 \
				libuuid libuuid-devel openssl \
				openssl-devel wget libgcc.i686 \
				gcc gcc-c++ cmake28 \
				cmake make glibc-devel.i686 \
				libgcc.i686 libstdc++-devel.i686; \
		yum -y clean \
				all

RUN git clone https://github.com/Zeex/pawn.git $PAWNCFOLDER; \
		cd $PAWNCFOLDER; \
		git checkout -b performance origin/performance; \
		mkdir build && cd build; \
		cmake ../source/compiler -DCMAKE_C_FLAGS=-m32 -DCMAKE_BUILD_TYPE=Release; \
		make; \
		mv libpawnc.so /lib/; \
		mv pawndisasm /bin/; \
		mv pawncc /bin/; \
		cd ~; \
		rm -rf $PAWNCFOLDER/;
```

Furthermore, these are the results using `time` pre-v3.10.3-performance.
```sh
real	1m39.890s
user	1m39.120s
sys	0m0.160s
```
```sh
real	1m52.975s
user	1m51.816s
sys	0m0.184s
```
```sh
real	1m51.449s
user	1m50.204s
sys	0m0.240s
```

These are the results using `time` with v3.10.3-performance.
```sh
real	0m9.082s
user	0m8.688s
sys	0m0.240s
```
```sh
real	0m10.315s
user	0m9.916s
sys	0m0.244s
```
```sh
real	0m9.977s
user	0m9.444s
sys	0m0.360s
```

Here is a screenshot of the continuous integration page of the latest builds, I think that the numbers speak for themselves.
chrome_2017-10-25_01-38-38

I will test if this pull request broke my gamemode, however I doubt it.

ziggi:
Tested with [Open-GTO](/Open-GTO/Open-GTO) gamemode and everything works fine. Vote for merge it! :boom:

Results on my i5-7500:
- Old time: 19.6s
- New time: 2.7s

7x speed improvement, this is awesome!:clap:

spacemud:
This version seems to omit the preceding character in strings when the line continuation operator is used:

```pawn
#include <a_samp>

main()
{
    new
        string[64] = 
        "\
        this is a \
        multi line \
        string \
        yay \
        ";

    print(string);

    return 1;
}
```

On 3.10.2 this gives:
```
this is a multi line string yay
```
...as expected.
On 3.10.3:
```
this is amulti linestringyay
```

I noticed this after getting errors from the MySQL plugin - all my multi line queries were getting mangled!

Zeex:
@spacemud I think it's because another patch that I did in 3.10.3, i.e. no this PR. I'll look into it.

Jeroen52:
Partay.

viruxe:
It's working with Scavenge and Survive. But still it's a 6sec diff from the russian one.

spacemud:
This was reported earlier and fixed in a later commit. Just waiting on a release with the fix.

PR #201, opened 2017-10-24
Slightly improve 'dumpn' write performance

Daniel-Cortez:


Zeex:
I'm having kind of a hard time understanding the encoding method used in `write_encoded` (even after staring at it for a couple of hours). Can you clarify why you're putting the bytes in reverse order in `encode_cell`? Is there any advantage in doing so?

I've tried running a script compiled with `#pragma compress 1` (which should trigger the use of compact encoding) and it seems that it's either encoding something badly or there is a hidden bug in the AMX that makes my server crash when it attempts to decode the data. For example, this script causes a crash:

```pawn
#pragma compress 1

#include <a_samp>

new x[10000] = {0xdeadbeef, ...};

main() {
	print("hello");
	printf("%x", x[0]);
}

```


Also when compiling without this patch (current version from master) I get a warning:

```
warning 232: output file is written, but with compact encoding disabled
```

but there is no such warning when compiling with the optimized encoding method. The corresponding check is located here:

https://github.com/Zeex/pawn/blob/c2a92d9bf320a87285dd0f107e9b7043ac065ced/source/compiler/sc6.c#L272-L273

AFAICT it's triggered when the output size is greater the input size by `AMX_COMPACTMARGIN-2`.

Daniel-Cortez:
> Can you clarify why you're putting the bytes in reverse order in encode_cell? Is there any advantage in doing so?

"Putting the bytes in reverse order" is what the original code does.
https://github.com/Zeex/pawn/blob/c2a92d9bf320a87285dd0f107e9b7043ac065ced/source/compiler/sc6.c#L248
For example, if a decoded value takes up 5 bytes, the bytes are saved into buffer `t` in the following order: 4-3-2-1-0. Then in a while loop it iterates down from 4 to 0 and writes the decoded bytes with `sc_writebin` as 0-1-2-3-4.
I have no idea why it was needed to save the bytes into the buffer in reverse order in the first place.

Anyway, what I did is reversed the order (in other words, made the compiler put the bytes in direct order) in order to be able to use `sc_writebin` for the whole encoded value only once instead of calling it multiple times for each decoded byte.

Daniel-Cortez:
> it seems that it's either encoding something badly or there is a hidden bug in the AMX that makes my server crash when it attempts to decode the data
Thanks for pointing it out, I miscalculated `bytes_in` and `bytes_out` in `write_decoded_n()`. I've commited a fix so it shouldn't happen now.

PR #204, opened 2017-10-26
 #pragma option 

Y-Less:
```pawn
#pragma option
```

Allows you to specify command-line options within a script.  Some don't work, some already have other pragma equivalents, but some can't be done yet.  Could actually replace a huge number of other pragmas:

```pawn
#pragma tabsize 4
#pragma option -t4

#pragma ctrlchar \
#pragma option -\

#pragma semicolon 1
#pragma option -;+

#pragma warning disable 202
#pragma option -w202

#pragma compat 1
#pragma option -Z+

#define HELLO 42
#pragma option HELLO=42

#pragma codepage 1424
#pragma option -c1424

// NO EQUIVALENT
#pragma option -(+

// NO EQUIVALENT
#pragma option -l

// NO EQUIVALENT
#pragma option -E

// NO EQUIVALENT
#pragma option -a

// etc...
```

Certain options are disabled in this manner, because they would alter details too fundamental, such as input and output files, that can't be changed after you already started reading said files:

```
-e
-o
-p
-r
```

Note that it will only read up to about 32 characters from the option, since it isn't really designed for long ones.  Speaking of which, a lot of the existing `pragma` code uses `i<sizeof name` and `name[i]='\0'`, which is technically a buffer overflow, but as far as I can see only ever writes an extra zero in to a subsequently unused variable, so is entirely safe.

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -1272,7 +1272,8 @@ static void parseoptions(int argc,char **argv,char *oname,char *ename,char *pnam
 
 void parsesingleoption(char *argv)
 {
-  parseoptions(1, &argv, NULL, NULL, NULL, NULL, NULL);
+  char *args[2] = { 0, argv };
+  parseoptions(2, args, NULL, NULL, NULL, NULL, NULL);
 }
 
 #if !defined SC_LIGHT
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -1053,8 +1053,7 @@ static void parseoptions(int argc,char **argv,char *oname,char *ename,char *pnam
         #endif
         break;
       case 'c':
-        if (codepage)
-          strlcpy(codepage,option_value(ptr),MAXCODEPAGE);  /* set name of codepage */
+        strlcpy(codepage,option_value(ptr),MAXCODEPAGE);  /* set name of codepage */
         break;
       case 'D':                 /* set active directory */
         ptr=option_value(ptr);
@@ -1272,8 +1271,14 @@ static void parseoptions(int argc,char **argv,char *oname,char *ename,char *pnam
 
 void parsesingleoption(char *argv)
 {
+  /* argv[0] is the program, which we don't need here */
   char *args[2] = { 0, argv };
-  parseoptions(2, args, NULL, NULL, NULL, NULL, NULL);
+  char codepage[MAXCODEPAGE+1] = { 0 };
+  codepage[0] = '\0';
+  parseoptions(2, args, NULL, NULL, NULL, NULL, codepage);
+  /* need explicit support for codepages */
+  if (codepage[0] && !cp_set(codepage))
+    error(108);         /* codepage mapping file not found */
 }
 
 #if !defined SC_LIGHT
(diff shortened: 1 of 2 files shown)


Romz24:
Good idea and implementation.

Jeroen52:
How can this be used to specify a variable through command-line parameters?

I'm using multiple `sed`s to pass environment variables to the gamemode

Y-Less:
This can't, but the normal command-line already can.

Jeroen52:
@Y-Less Thank you! I already throught I read that somewhere but I couldn't find it when I tried to find it later.

Now I can finally throw away my hack-ish `sed` prompts.

Not even joking, it looks like this.

```sh
   sed -i "s/#define localmode 1/#define localmode $LOCALMODE/" $CI_PROJECT_DIR/gamemodes/SFCRRPG.pwn
   sed -i "s/#define SQL_SERVER		\"127.0.0.1\"/#define SQL_SERVER		\"$SQL_SERVER\"/" $CI_PROJECT_DIR/include/mysql.inc
   sed -i "s/#define SQL_USER		\"D3V1Lx\"/#define SQL_USER		\"$SQL_USER\"/" $CI_PROJECT_DIR/include/mysql.inc
   sed -i "s/#define SQL_DB			\"CNRUSERS\"/#define SQL_DB			\"$SQL_DB\"/" $CI_PROJECT_DIR/include/mysql.inc
   sed -i "s/#define SQL_PASS 		\"johncena\"/#define SQL_PASS 		\"$SQL_PASS\"/" $CI_PROJECT_DIR/include/mysql.inc
```

Even looking back at that, I could've easily made multiple replacements with only one sed. But I have only learnt that recently.

Y-Less:
While that is correct, you shouldn't be building usernames and passwords in to scripts like that.  Loading those is now one of the best remaining uses for an INI system to store your credentials for connecting to a database.

Jeroen52:
You're right, however the binaries containing the username and password are automatically compiled and sent to the server running the gamemode. The credentials itself are saved in GitLab.

And if someone were to obtain the AMX, I could just revoke the credentials and issue new ones.

Y-Less:
You shouldn't be storing credentials in git either!

Jeroen52:
Not in Git, in GitLab as a protected secret variable.

I know very well that you shouldn't store secrets in Git.

Y-Less:
Fixed a bug and added `-c` support.

afwn90cj93201nixr2e1re:
@Y-Less could i ask something?

https://github.com/pawn-lang/compiler/blob/f71d55d380937eba821b624c10afa2cdbcce2b7c/source/compiler/sc1.c#L1056
https://github.com/pawn-lang/compiler/blob/f71d55d380937eba821b624c10afa2cdbcce2b7c/source/compiler/sc1.c#L1099

**why there's no checks** against nullptr on few cases)-i and -c, i understand **that maybe missing -c checks expected** and blablabla, coz i really don't understand what's -c (codepage) stays for at this cases, but what about include dir? Is that expected?



https://github.com/pawn-lang/compiler/blob/134ad7a836c581546665340aedb59efd4636e269/source/compiler/sc1.c#L1576

8 means sCHARBITS, right? If yes, could you just change it too.

afwn90cj93201nixr2e1re:
@Y-Less, @YashasSamaga. @Zeex

afwn90cj93201nixr2e1re:
So, dont answer, im gonna extend it little bit more.

Issue #155, opened 2017-02-25
Tokenisation operator - ##


Y-Less:
I don't know why I've never thought to ask for this before.  Frankly I don't know why I didn't do it in the first place when I added the `#` operator all those years ago...  But I didn't.  Anyway, I was just writing up some text on my new library for getting round one of the major limitations of the pre-processor, that is, things like:

```pawn
#define COMMAND_NAME throw
CMD:COMMAND_NAME(playerid, params[])
{
}
```

Not working as many would expect.  I figured out one way of doing it, but that way would only work for very special macros that looked like:

```pawn
#define COMMAND_NAME COMMAND_NAME_()
#define COMMAND_NAME_ MACRO(throw)
CMD:COMMAND_NAME(playerid, params[])
{
}
```

And then wouldn't work for things NOT using macros, so this would start to fail:

```pawn
CMD:throw(playerid, params[])
{
}
```

I was just writing about how the best solution would really be `##`, to enable this, but it doesn't exist:

```pawn
#define CMD:%0(%1) \
	forward cmd_##%0(%1); \
	public  cmd_##%0(%1)
```

Then I thought - WHY doesn't it exist!?  It would solve SO MANY problems!

Y-Less:
It wouldn't solve all problems, since the `cmd_##%0` would not be evaluated as a pre-processor token where `cmd_%0` would, and that is sometimes desired.  But you can pick-and-choose which to use.

Y-Less:
I'm sure there are still cases where you may not know in advance which one you need, but it would still enable more solutions.  Better yet, have the best of both worlds - make the pre-processor aware of both variations, such that it sees:

```pawn
cmd_##MACRO
```

First as `cmd_MACRO` then `cmd_` followed by `MACRO`, i.e. tries to evaluate it both ways and thus allow replacements for whole symbols and part-symbols (normally a bad idea, but essentially explicitly requested here).

Y-Less:
Note: IMHO the order should be whole symbol first, THEN parts, with the `##` being invisible to the first version but maybe not the second:

```pawn
#define cmd_MACRO() 10 // Matches.
#define cmd_##MACRO() 20 // Would match if the first hadn't.
#define cmd_%0() // Would also match, just because of existing rules.

cmd_##MACRO()
```

No idea what `%0` would be in that third example though - the obvious answer is `##MACRO`, but should the `##` be skipped?  Or should this be valid if you use `%0` as another input:

```pawn
[...]

PR #205, opened 2017-10-27
Tokenisation operator

VVWVV:
#155
Example:
```pawn
#define foo(%1,%2) %1##%2

main() {
    new barbaz;
    printf("%d", foo(bar, baz));
}
```

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1830,15 +1830,14 @@ static void substallpatterns(unsigned char *line,int buffersize)
     } /* if */
   } /* while */
 }
-#endif
 
 static void ppconcat(const char *line)
 {
-  register const char *c, *p, *b;
+  const char *c,*p,*b;
   for (c=line; *c!='\0'; ++c) {
     if (is_startstring(c))
       c=skipstring(c);
-    if(*c=='#' && *(c+1)=='#') {
+    if (*c=='#' && *(c+1)=='#') {
       p=c-1;
       b=c+2;
       while (*p==' ')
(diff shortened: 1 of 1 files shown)


Zeex:
This patch needs to check if the `##` is inside a string literal because:

```pawn
print("##");
```

^ will print `(null)`.

Crytrax:
This patch breaks YSI. The following is the compiler's output.
```
Dependencies\CoreUtils\YSI\Init\..\YSI_Server\..\YSI_Data\..\YSI_Internal\..\YSI_Core\..\YSI_Coding\y_va.inc(112) : error 027: invalid character constant
```
Can anyone else confirm that this is not just some mistake on my end?

Dayvison:
This commit 0786aed is breaking

Crytrax:
Commit https://github.com/Zeex/pawn/pull/205/commits/f59b285e171eaf00a589df072597ded58d5be5bb fixes this. Thanks @VVWVV

Y-Less:
One thing to be aware of, as I've seen this a few times:

```pawn
#define COLOR_RED #FF0000

SendClientMessage(playerid, COLOR_GREEN, "Hello {"#COLOR_RED"} world");
```

Not sure what would happen there.  But it shouldn't be too hard to avoid by skipping backtracking too much when you do symbol replacements.

Zeex:
If you're going to work on this and re-submit, please also include a test that covers as much test cases as you can think of. For example, `##` inside macros, outisde of macros, with spaces around and without, insides strings and outside, and so on.

We're starting to add compile and crash tests for new issues (and hopefully runtime tests soon). See [source/compiler/tests](https://github.com/Southclaws/pawn/tree/master/source/compiler/tests) and [source/compiler/tests/CMakeLists.txt](https://github.com/Southclaws/pawn/blob/master/source/compiler/tests/CMakeLists.txt) for examples.

PR #215, opened 2017-12-02
Do not allow to redefine labels

Daniel-Cortez:
```Pawn
test()
{
my_label:
    // ...
    if (/* ... */)
        goto my_label; // the jump is made to the place of the 2'nd
                       // definition of 'my_label' below
    // ...
my_label: // before this PR there are no errors/warnings at label redefinition
}
```
After this PR the compiler issues error 021 ('symbol already defined: "my_label"').
The same bug has been fixed in Pawn 4.0 a year ago.


Zeex:
That makes sense. Thanks.

PR #222, opened 2017-12-28
BUG fix in findfunction

Terchik:


Zeex:
Thanks for the fix, but this project covers only the compiler part of the Pawn toolkit, not runtime.

Issue #220, opened 2017-12-22
Multiple large initialised globals crash in the wrong order


Y-Less:
Crashes:

```pawn
new a[1000][500] = { { 0, 1, ... }, ... };
new b[500][2000] = { { 0, 1, ... }, ... };
```

Doesn't crash:

```pawn
new b[500][2000] = { { 0, 1, ... }, ... };
new a[1000][500] = { { 0, 1, ... }, ... };
```

I'm guessing it has something to do with allocating memory, since the crash doesn't happen if the total size of the second array is <= the first.

YashasSamaga:
https://image.prntscr.com/image/TTe7rjEESRC-jSBUMzp4FA.png

I can't understand why that's even happening but there's the problem.

YashasSamaga:
Replaced the pointers with indexes and it's working.

https://github.com/YashasSamaga/pawn/commit/538954b48345c7635576bfb0e724791de7f07c2d

PR #225, opened 2017-12-31
replaces dangling pointers in initarray with indexes

YashasSamaga:
fix for #220 

In `initarray`, `prev1` and `prev2` are dangling pointers. When assigned, they point to an address of an item in the literal queue. The compiler checks (`litadd`) the literal queue size before adding an element. If the size isn't sufficient to hold another element, it reallocates the literal queue to accommodate more items. When the reallocation happens, the previous references (`prev1` and `prev2`) to the elements of the literal queue are invalidated. De-referencing the broken pointers will cause undefined behavior.

This commit replaces the pointers with indexes. The indexes are invariant to the reallocation.


Zeex:
Nice catch 👍

Issue #172, opened 2017-06-04
Warnings for meaningless code


Y-Less:
```pawn
Func(const &var)
{
}
```

If you modify the variable in the function you will get an error.  If you don't, why is it a reference?  The combination is currently allowed but silly.  Similarly:

```pawn
Func(const ...)
{
}
```

That is allowed by the compiler, but is probably even worse - the `const` is totally ignored.  You can pass `const` variables when it isn't there, or modify variables via `setarg` when it is there.  I don't actually know what the best solution to that would be because obviously `setarg` is pretty much a total compiler bypass.

Probably just warnings for those `const`s that they are meaningless.

YashasSamaga:
Would this be meaningless?

`static const k;`

The behavior of the above is well defined. The compiler initializes the value to zero.

Y-Less:
How is that related?  I'm talking about the use of both `const` - value can't be modified, and `&` - source value will be modified, together.  Basically `const &` says "this value can't be modified, and any modifications are propagated".

YashasSamaga:
I thought of using this issue to discuss other kinds of meaningless code instead of having different issues for each.

Y-Less:
I actually found a use for `const &` since writing this:

```pawn
ModifyConst(const & var, val)
{
    setarg(0, val);
}
```

Or course, since I just argued against not modifying the language for the sake of maintaining compatibility with awkward workaround code, in the rare case where this is desired the warning can be disabled and re-enabled:

https://github.com/Open-GTO/sa-mp-fixes/blob/c4eda3041f2935dad7ddae7c8a3892a29781974c/fixes.inc#L12015-L12026

On your issue.  Valueless variables are very well defined and widely as `0` in PAWN everywhere.  I don't think `static const x;` in any way goes against that.

Y-Less:
As for what the warning should be, something like "incompatible combination of class specifiers".  Error 42 is "invalid combination of class specifiers", this isn't quite invalid, I don't even like the word "incompatible", but something alliteratively similar but with slightly less connotation of being totally wrong.  "Unlikely", "ignored", "meaningless", or something else like that.

YashasSamaga:
```
func1(const ...) { }
func2(const& x) { }
func3(const x) { }
func4(...) { }
func5(&x) { }
func6(x) { }
main ()
{
	new a = 5;
	func1(a);
	func2(a);
	func3(a);
	func4(a);
	func5(a);
	func6(a);
}
```

```
[...]

PR #246, opened 2018-01-07
warning for meaningless class combinations

YashasSamaga:
#172 

1. Adds a generic new warning which is triggered when meaningless combination of class specifiers are used.

2. This commit adds code to trigger a warning for the following two cases:
- constant reference
- - not meaningful as as references always point to cells in PAWN (unlike C++ where it would help prevent costly copies while guaranteeing to not modify the object) which makes const reference as good as pass-by-value
- constant variable arguments
- - for similar reasons as in the previous case

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -3867,6 +3867,8 @@ static int declargs(symbol *sym,int chkshadow)
       case '&':
         if (ident!=iVARIABLE || numtags>0)
           error(1,"-identifier-","&");
+        if (fconst)
+          error(238, "const reference"); /* meaningless combination of class specifiers */
         ident=iREFERENCE;
         break;
       case tCONST:
@@ -3941,6 +3943,8 @@ static int declargs(symbol *sym,int chkshadow)
       case tELLIPS:
         if (ident!=iVARIABLE)
           error(10);                    /* illegal function or declaration */
+        if (fconst)
+          error(238, "const variable arguments"); /* meaningless combination of class specifiers */
         if (numtags==0)
           tags[numtags++]=0;            /* default tag */
         if ((sym->usage & uPROTOTYPED)==0) {
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -191,7 +191,8 @@ static char *warnmsg[] = {
 /*234*/  "function is deprecated (symbol \"%s\") %s\n",
 /*235*/  "public function lacks forward declaration (symbol \"%s\")\n",
 /*236*/  "unknown parameter in substitution (incorrect #define pattern)\n",
-/*237*/  "user warning: %s\n"
+/*237*/  "user warning: %s\n",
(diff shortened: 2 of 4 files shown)


Zeex:
Can you also add a test? It can be a simple compilation test that checks if the warnings are printed correctly (see [source/compiler/tests](https://github.com/Southclaws/pawn/tree/master/source/compiler/tests)).. Maybe optionally check valid combinations of specifiers so that someone some future change doesn't break them by accident.

YashasSamaga:
I have added the tests in a separate branch to prevent this PR from polluting the commit history. I'll move them here as a single commit once I'm sure that I'm doing this right.

Test script: https://github.com/YashasSamaga/pawn/commit/52b97f6ec7bd2a4da809a05f588f5106f947a151

CMakeLists.txt change: https://github.com/YashasSamaga/pawn/commit/b3ed64b44eaf40627c8018df8e440f692aaa5660

Is that how it is supposed to be?

1. is `meaningless_class_specifiers_warnings_246.pwn` over-verbose?
2. what does `gh` mean?
3. suggestions?

Y-Less:
Could you add a test for warning pop/disable/push around it please?  Just to ensure it will work in the very rare case where it is desired.  Thanks.

Zeex:
`gh` means GitHub, so `gh_246` would be "GitHub issue 246". You can name it whatever you want, just make sure that the issue number is included in the name so it's easy to look up the corresponding issue.

YashasSamaga:
@Y-Less Those directives work independently of the commits in this PR. The commits in this PR use the `error` function to trigger the warnings. Whether to trigger the warning or not (because it is enabled or disabled, or if the compiler is in first pass or second pass, or whatever the reason maybe) is decided by the `error` function. Therefore, I think the tests you suggested are not really relevant to this PR.

As one could expect, the directives will work with this warning too:

```
#pragma warning push
#pragma warning disable 238
f1(const &v) { }
#pragma warning pop
f2(const ...) { }
f3(const v) {
	#pragma unused v
}
f4(...) { }
f5(v) {
	#pragma unused v
}
f6(&v) { }

main() {
	new a;
	f1(a);
	f2(a);
	f3(a);
	f4(a);
	f5(a);
	f6(a);
}
```

**Output:**
```
warning 238: meaningless combination of class specifiers (const variable arguments)
```

Zeex:
Test name should refer to the bug/issue number, not the PR number since it tests that there is no bug, not that the PR works.  If a PR adds some new feature I wouldn't add its number to the name at all. 

It could be useful to know where is the original bug report on GitHub that a test addresses, but not as much useful to know which PR implemented the fix (or a feature). It could also be multiple PRs that address the same problem. Well, I don't know, it's kind of hard to explain, maybe I'm overcomplicating things and everybody should just name tests whatever they want.... I haven't done a lot of automated tests in the past.

Just use a descriptive name that makes sense and is short enough to fit on screen.

PR #255, opened 2018-01-13
Memoize the last node in constvalue lists

Daniel-Cortez:
The first node in constvalue lists is always treated differently than the others, only its `next` field is used, so it should be safe to reporpose a part of unused space to memoize the address of the last node in the list. This should speed up the addition of new nodes to the end of the list.

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -113,6 +113,10 @@ typedef struct s_constvalue {
                          * tag for enumeration lists */
 } constvalue;
 
+typedef struct s_constvalue_root {
+  constvalue *next,*last;
+} constvalue_root;
+
 /*  Symbol table format
  *
  *  The symbol name read from the input file is stored in "name", the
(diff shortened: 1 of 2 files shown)


Daniel-Cortez:
Tested compiling "modelsizes.inc" (#285)

```Pawn
#include "../include/modelsizes.inc"
#pragma unused MODELS_gColRadius, MODELS_gColOffset, MODELS_gColAABB
main(){}
```

On my machine the compile time has decreased from ~13.7 to ~0.95 seconds.

maddinat0r:
Your changes will make the compiler crash if someone in the future decides to actually write something into the `name` array of a root `constvalue` node. You should at least add a comment in the code describing the behavior you depend on, and explicitly warn about not writing anything into the root node.

Daniel-Cortez:
@maddinat0r Well yeah, now that I look at the changes I've made in this PR I think the current implementation is pretty hackish. Just adding comments won't really help, it rather needs a complete rewrite and that's exactly what I'm doing as of writing this.

Daniel-Cortez:
Ok, so this time instead of just being casted from `constvalue` to `constvalue_root` all root nodes are explicitly defined as of type `constvalue_root`. This way the code should be more readable since now it's easier to tell the root nodes from the regular ones.

PR #269, opened 2018-01-19
Optimize string copying in rebuffer() and stgwrite()

Daniel-Cortez:
In this patch I've replaced loops with `CHECK_STGPIPE`/`CHECK_STGBUFFER` checks at every iteration with a single buffer growth check and a `memcpy` call.
This doesn't do much performance-wise, but I think it's still good to have.

Also it seems that the same kind of optimisation has already been applied in `stgwrite` within the `else` branch, but for some reason not within the `if (staging)` branch.

### DIFF
--- a/source/compiler/sc7.c
+++ b/source/compiler/sc7.c
@@ -1332,14 +1332,12 @@ SC_FUNC void stgmark(char mark)
 static int rebuffer(char *str)
 {
   if (sc_status==statWRITE) {
+    int st_len=strlen(str);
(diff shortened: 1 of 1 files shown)


maddinat0r:
> Also it seems that the same kind of optimisation has already been applied in stgwrite within the else branch, but for some reason not within the if (staging) branch.

That was actually my doing! The unoptimized code in the `else` branch was overall a pretty big bottleneck back then, that's why I optimized only that part.

PR #270, opened 2018-01-20
trigger unused symbol warning for references

YashasSamaga:
If a reference argument is not read within the function, the compiler does not trigger an unused symbol warning.

This commit modifies the compiler to trigger warnings in such cases.

I have spent some time checking through every reference that was made to the `uREAD` flag in the entire codebase. As far as I could see, there would be no side-effects of this commit.

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -4134,8 +4134,6 @@ static void doarg(char *name,int ident,int offset,int tags[],int numtags,
     assert(numtags>0);
     argsym=addvariable(name,offset,ident,sLOCAL,tags[0],
                        arg->dim,arg->numdim,arg->idxtag,0);
-    if (ident==iREFERENCE)
-      argsym->usage|=uREAD;     /* because references are passed back */
     if (fpublic)
       argsym->usage|=uREAD;     /* arguments of public functions are always "used" */
     if (fconst)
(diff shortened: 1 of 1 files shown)


Zeex:
References are usually used for output arguments, i.e. you only write to them. For example, `GetPlayerPos()`. Perhaps it would make more sense to warn if parameters passed by reference are not written to within a function body? I imagine it could be pretty hard to detect since functions can call other functions and so on.

YashasSamaga:
If nested function calls keep passing the reference around and if the original reference is not written to, then there will be a warning at the deepest function in the call hierarchy.

This is probably better than triggering a warning in every function which passes a reference to other functions. Just one warning will be triggered for the entire call hierarchy instead of a warning in each function of the call hierarchy.

With this commit references which are not read nor written will trigger 203. Your suggestion is that a warning has to be triggered if the reference is not written to because the purpose of passing the reference is lost?

But a lot of codebases use assembly to obtain the address of the symbol. While the compiler is smart enough to mark `uREAD` (it actually marks it wrongly, i.e: marks as `uREAD` even if it was actually a write operation) but isn't smart enough to mark `uWRITTEN` flag. This would cause false warnings to be triggered that a reference wasn't written to even if it might've been modified through assembly or calls to functions made from assembly.

Issue #273, opened 2018-01-27
Bad function adress in __emit


samphunter:
```C
#include <a_samp>

public OnFilterScriptInit() {
	new Func:a, Func:b;
	
	__emit const.pri TestFunc;
	__emit stor.S.pri a;
	
	#emit CONST.pri  TestFunc
	#emit STOR.S.pri b
	
	printf("%x == %x", _:a, _:b);
        return 1;
}

TestFunc() {
	print("Success!");
	return 1;
}
```

This produces `9D4 == B8`. I assume this is because __emit takes absolute adress while #emit takes relative, but this should be fixed so that both use relative adress.

YashasSamaga:
What is `TestFunc`? I suspect that your test code doesn't even compile.

samphunter:
@YashasSamaga sorry, edited to icnlude TestFunc(). Also it does not seem to be relative adress, as I am trying to account for it and am failing.

YashasSamaga:
Ah, I think that's happening because `TestFunc` is declared later.

samphunter:
@YashasSamaga Yes, putting TestFunc before the emit seems to fix the problem. Though I would say it would be preferable, that the compiler could handle this entirely, just like the #emit version can. Undefined symbol would still be annoying for many usecases, such as my attempt to use function pointers as arguments to other functions.

YashasSamaga:
Actually I was wrong about undefined symbol. That error is applicable to variables only.

The `__emit` syntax seems to be providing the wrong address.

Zeex:
Closed because fixed by #279.

PR #279, opened 2018-01-31
Another bunch of fixes/improvements for 'emit'

Daniel-Cortez:
Changes:<ul>
<li>Added an ability to use labels in arguments of type 'any' (<code>const.pri/alt</code>, <code>add.c</code> etc.)
Example:

```Pawn
main()
{
lbl1:
	emit const.pri lbl1;
	emit sctrl 6;
}
```

Assebmler output (<code>-a</code>):

```
	proc	; main
	; line 2
	; line 3
l.0		; c
	lctrl 5
	sctrl 4
	; line 4
	const.pri l.0
	;$exp
	zero.pri
	retn
```

Disassembler (<code>pawndisasm</code>) output:

```
00000008  proc 
0000000c  lctrl 00000005
00000014  sctrl 00000004
0000001c  const.pri 0000000c
00000024  zero.pri 
00000028  retn 
```

If the label is used before the point of its implementation, it should be prefixed with the ':' symbol.

```Pawn
    //emit const.pri label1; // 'error 017: undefined symbol: "label1"'

    emit const.pri :label1; // by specifying ':' before the symbol name
label1:                     // we let the compiler know this is a label,
                            // not an undefined variable or constant,
                            // so it can be expected to be implemented later

    emit const.pri label1;  // not specifying the prefix would still work,
                            // but only if the label is implemented
                            // before the point of its use

    emit jump :label2; // specifying ':' in jump instructions is optional,
    emit jump label2;  // even if the label is implemented after the point of its use
label2:
```
</li>

<li>Added new argument types:<ul>
<li>Integer value<br/>
Any integer value from <code>cellmin</code> to <code>cellmax</code> (<code>-2147483648..2147483647</code> in case of SA-MP).<br/>
Used in: <code>jrel</code>, <code>casetbl</code> (2'nd argument), <code>cmps</code>, <code>movs</code>, <code>fill</code>, <code>halt</code>, <code>bounds</code>.
</li>
<li>Non-negative integer<br/>
Subset of 'Integer value' with valid values ranging from <code>0</code> to <code>cellmax</code>.<br/>
Used in: <code>casetbl</code>, <code>cmps</code>, <code>movs</code>, <code>fill</code>, <code>halt</code>, <code>bounds</code>.

```Pawn
emit movs 128; // ok
emit movs (MAX_PLAYERS * cellbits / charbits); // expressions are allowed

emit halt -1; // error 001: expected token: "-nonnegative integer-", but found "-1"
emit movs 0xFFFFFFFE; // error 001: expected token: "-nonnegative integer-", but found "-2"
```
</li>
<li>Shift<br/>
Subset of 'Integer value' with valid values ranging from <code>0</code> to <code>cellbits - 1</code> (which is <code>31</code> for SA-MP).<br/>
NOTE: Expressions are NOT allowed.<br/>
Used in: <code>shl.c.pri/alt</code>, <code>shr.c.pri/alt</code>, <code>idxaddr.b</code>, <code>lidx.b</code>.

```Pawn
emit shl.c.pri 0; // ok
emit shr.c.alt 31; // this is ok too

const SHIFT = 16;
emit shl.c.pri SHIFT; // named constants can be used as well, of course

emit shr.c.pri 32; // error 050: invalid range
```
</li>
<li>Alignment<br/>
Subset of 'Integer value' with valid values ranging from <code>0</code> to <code>sizeof(cell) - 1</code> (<code>0..3</code> for SA-MP).<br/>
NOTE: Expressions are NOT allowed.<br/>
Used in: <code>align.pri/alt</code>.
</li>
</ul>
</li>

<li>Disallowed the use of the '-' sign with function names in arguments of type 'any'.</li>

<li>Fixed invalid function address when the function is implemented after the point of its use with <code>emit</code> (#273).</li>

<li>Now for arguments of type "any" (e.g. <code>const.pri/alt</code>, <code>add.c</code>, <code>eq.c.pri/alt</code>) in case of error the compiler says it <b>actually</b> expected an argument of type "any value" (previously it told it expected a "numeric value" instead, even though variables and functions worked too).
</li>
</ul>


Daniel-Cortez:
Things I'm still not sure about:<ul>
<li>Is 'nonnegative number' a good name for an argument type? I don't know English very well, so I can't really tell.
</li>

<li>Currently we can use labels in arguments of type 'any', but only if the label is implemented before the point of its use with `emit`.
  <br/>It's possible to use labels implemented after the point of their use for jump instructions (see `emit_param_label()`), because there I know the instruction argument is going to be a label, so I can use `fetchlab()` to make the compiler expect the label to be implemented later, but I can't do the same in `emit_param_any()` since there's no way to know if the currently undefined symbol is going to be a label (so it could be expected to be implemented later) or of any other type (in this case the compiler should issue an 'undefined symbol' error).

  In order to solve this ambiguity, we could use an optional prefix before the label name, like this:

```Pawn
    emit const.pri :label; // by specifying ':' before the symbol name
label:                     // we let the compiler know this is a label

    emit const.pri label;  // not specifying the prefix would still work,
                           // but only if the label is implemented
                           // before the point of its use
```

  but would this kind of solution be good at all? And if so, what prefix would be preferable? (I'd personally prefer the ':' one since it's used for label definition as well.)
</li>
</ul>

samphunter:
> Is 'nonnegative number' a good name for an argument type?

Either that or unsigned integer/natural number. Depends on how technical you want it to be and whether you prefer the IT or mathematical term.

> but would this kind of solution be good at all? And if so, which prefix would be preferable? (I'd personally prefer the : one since it's also used as a postfix for label definition, but maybe there are better options.)

Sounds good to me and yes, I think we should have a way to use a label defined after the code, since those are often needed. I also like the `:` syntax. Can't imagine any other prefix that would fit so well.

Daniel-Cortez:
> unsigned integer

I didn't use this term because it also implies values from `0x80000000` to `0xffffffff` (provided we have 32-bit cells) which is misleading since the VM would treat them as signed (and therefore negative) values anyway.
Though I have to thank you for making me double-check the latter. I looked at the code in `amx_Exec()` and found out the `bounds` handler actually [treats](https://github.com/pawn-lang/compiler/blob/master/source/amx/amx.c#L2716-L2717) the argument as a `ucell` (unsigned cell), so I guess I'll have to remove the sign check for this opcode in `emit`.

> natural number

This would also imply floating-point numbers, which isn't quite right either.


> Sounds good to me and yes, I think we should have a way to use a label defined after the code, since those are often needed. I also like the `:` syntax. Can't imagine any other prefix that would fit so well.

Thanks for the feedback. I believe it's worth to mention this idea partially belongs to @VVWVV as well - he initially planned to require the `:` sign after (not before) any label name back when he just started implementing the `emit` operator (it was called `_asm` back then), but the idea was scrapped afterwards.

samphunter:
> This would also imply floating-point numbers, which isn't quite right either.

That is actually incorrect, natural numbers implies whole numbers, often from 1 up. So the only problem with that is it does not imply 0 (but some mathematicians use it that way, they can't really agree whether 0 should be included).

Y-Less:
I'd say non-negative numbers.  While natural numbers may be mathematically accurate, it isn't a "natural" term for non pure mathematicians, which is most of the target audience.  Needing a degree in maths just to understand warning messages isn't great.

samphunter:
> I'd say non-negative numbers.

But that sound as if it meant real non-negative numbers.

I would go with `non-negative integers` or `non-negative whole numbers` depending on how simple you want it.

Daniel-Cortez:
Well, speaking of simplicity, initially I was going for "nonnegative integer value", which corresponded with other existing type names such as "integer value", "rational value" and "numeric value" (see the `sc_tokens` array in sc2.c), but I thought the name sounded too complicated, so I shortened it to just "nonnegative integer". I can change it back though, I'm just not sure what's more important in this case: simplicity or consistency with the other type names.

PR #282, opened 2018-02-03
Documentation improvement ▼

Dayvison:
- [x] Add `location` tag `<location file="" startline="" endline="">`
- [x] Fix `transition` tag
- [ ] Escape special characters in documentation
- [x] Fix enumerators documentation
- [x] Fix constant documentation
- [x] Add documentation for enumerators items
- [x] Add new atributes for member
	+ [x] const
	+ [x] deprecated
	+ [x] entry
	+ [x] enum
	+ [x] enum_field
	+ [x] forward
	+ [x] missing
	+ [x] naked
	+ [x] native
	+ [x] pre_defined
	+ [x] prototyped
	+ [x] public
	+ [x] returns
	+ [ ] return_type
	+ [x] static
	+ [x] stock
	+ [x] unused
- [x] Update `pawndoc.xsl` to support new features
- [ ] Fix a bug, when you documentate one prototype without keyword `forward` the documentation go to global

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -155,9 +155,10 @@ typedef struct s_symbol {
     } array;
   } dim;                /* for 'dimension', both functions and arrays */
   constvalue *states;   /* list of state function/state variable ids + addresses */
-  int fnumber;          /* static global variables: file number in which the declaration is visible */
+  int fnumber;          /* file number where is declared */
   int lnumber;          /* line number (in the current source file) for the declaration */
-  int lnumber_end;      /* line number (in the current source file) for the end of declaration */
+  int lnumber_decl;     /* line number where starts the declaration (ignore prototype) */
+  int lnumber_end;      /* line number for the end of declaration */
   struct s_symbol **refer;  /* referrer list, functions that "use" this symbol */
   int numrefers;        /* number of entries in the referrer list */
   char *documentation;  /* optional documentation string */
@@ -194,6 +195,7 @@ typedef struct s_symbol {
  *        3     (uCONST) the variable is constant (may not be assigned to)
  *        4     (uPUBLIC) the variable is public
  *        6     (uSTOCK) the variable is discardable (without warning)
+ *        7     (uSTATIC) the variable is static
  *
  *  FUNCTION
  *  bits: 0     (uDEFINE) the function is defined ("implemented") in the source file
@@ -205,6 +207,7 @@ typedef struct s_symbol {
  *        6     (uSTOCK) the function is discardable (without warning)
  *        7     (uMISSING) the function is not implemented in this source file
  *        8     (uFORWARD) the function is explicitly forwardly declared
+ *        9     (uSTATIC) the function is static
  *
  *  CONSTANT
  *  bits: 0     (uDEFINE) the symbol is defined in the source file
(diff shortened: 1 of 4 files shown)


Dayvison:
I don't know what will be the tests for this.

I've change how the compiler check if symbol are static, but don't worry all work fine(Tested)

Y-Less:
I use several organisational tags, including `<section>`, `<subsection>`, and `<library>`.  It would be nice if these were properly hierarchical so that functions declared within one stayed within it.

Dayvison:
Can you give a example of usage?

source/compiler/sc.h
```diff
@@ -227,6 +232,7 @@ typedef struct s_symbol {
 #define uENUMFIELD  0x040
 #define uMISSING    0x080
 #define uFORWARD    0x100
+#define uSTATIC     0x200
```
YashasSamaga: There's already `sSTATIC` bit which is used to mark symbols as static. It's stored in the `vclass` member. Is there any particular reason why you decided not to use it and add a your own?

source/compiler/sc1.c
```diff
@@ -3400,7 +3454,7 @@ SC_FUNC void check_tagmismatch(int formaltag,int actualtag,int allowcoerce,int e
 {
   if (!matchtag(formaltag,actualtag,allowcoerce)) {
     constvalue *tagsym;
-    char formal_tagname[sNAMEMAX+3]="none (\"_\"),",actual_tagname[sNAMEMAX+2]="none (\"_\")"; /* two extra characters for quotes */  
+    char formal_tagname[sNAMEMAX+4]="none (\"_\"),",actual_tagname[sNAMEMAX+3]="none (\"_\")"; /* two extra characters for quotes */
```
YashasSamaga: Assume that `sNAMEMAX` is `32` (which it is in our compiler). An identifier can have at most 31 characters. The 32nd character will be used as a null terminator for the longest identifier.

With 31 significant characters of the tagname, two quotes and a comma (for `formal_tagname`), the maximum length turns out to be 31(identifier) + 2(quotes) + 1 (comma) + 1 (null terminator) = 35.

`sNAMEMAX + 3` = 35

For `actual_tagname`, it'll be one lesser as there is no need for comma.

I can't see why the length has been increased by one for both.

source/compiler/sc.h
Dayvison: what you suggest? i don't think will confuse

Dayvison:
@YashasSamaga @Y-Less The actual way is ignore all unused symbols, in my opinion even symbol are unused they should go to documentation, with `unused` attribute

source/compiler/sc.h
YashasSamaga: The information which could be obtained from `sSTATIC` can now be obtained from `uSTATIC` and `vclass` (by using `sLOCAL`, `sGLOBAL` and `sSTATEVAR`).

I was thinking of removing `sSTATIC` and use `sGLOBAL`/`sLOCAL`/`sSTATEVAR` in `vclass`. If both `uSTATIC` and `sLOCAL` is set, it implies what `sSTATIC` previously was.

But this would require a lot of editing and could potentially introduce new bugs.

Y-Less:
@YashasSamaga You've looked in to this.  Merge thoughts?

YashasSamaga:
@Dayvison can you please rebase onto dev?

PR #287, opened 2018-02-25
Back-ported some PAWN 4 natives and features.

Y-Less:

Backport natives and features from PAWN 4.

Fixes #

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

This takes the default natives from PAWN 4 and puts them in PAWN 3.

### DIFF
--- a/source/amx/amx.h
+++ b/source/amx/amx.h
@@ -27,54 +27,63 @@
 #include <stdlib.h>   /* for size_t */
 #include <limits.h>
 
+#if (defined __linux || defined __linux__) && !defined __LINUX__
+  #define __LINUX__
+#endif
 #if defined FREEBSD && !defined __FreeBSD__
   #define __FreeBSD__
 #endif
-#if defined LINUX || defined __FreeBSD__ || defined __OpenBSD__
+#if defined __LINUX__ || defined __FreeBSD__ || defined __OpenBSD__ || defined __APPLE__
   #include <sclinux.h>
 #endif
 
+#if defined __GNUC__
+ #define GCC_VERSION (__GNUC__ * 10000 \
+                               + __GNUC_MINOR__ * 100 \
+                               + __GNUC_PATCHLEVEL__)
+#endif
+
+#if !defined HAVE_STDINT_H
+  #if (defined __STDC_VERSION__ && __STDC_VERSION__ >= 199901L) \
+      || defined __GNUC__ || defined __LCC__ || defined __DMC__ \
+      || (defined __WATCOMC__ && __WATCOMC__ >= 1200)
+    #define HAVE_STDINT_H 1
+  #endif
+#endif
+#if !defined HAVE_INTTYPES_H
+  #if defined __FreeBSD__ || defined __APPLE__
+    #define HAVE_INTTYPES_H 1
+  #endif
+#endif
 #if defined HAVE_STDINT_H
   #include <stdint.h>
+#elif defined HAVE_INTTYPES_H
+  #include <inttypes.h>
 #else
-  #if defined __LCC__ || defined __DMC__ || defined LINUX || (defined __WATCOMC__ && __WATCOMC__ >= 1200)
-    #if defined HAVE_INTTYPES_H
-      #include <inttypes.h>
+  #if defined __MACH__
+    #include <ppc/types.h>
+  #endif
+  typedef short int           int16_t;
+  typedef unsigned short int  uint16_t;
+  #if defined SN_TARGET_PS2
+    typedef int               int32_t;
+    typedef unsigned int      uint32_t;
+  #else
+    typedef long int          int32_t;
+    typedef unsigned long int uint32_t;
+  #endif
+  #if defined __WIN32__ || defined _WIN32 || defined WIN32
+    typedef __int64	          int64_t;
+    typedef unsigned __int64  uint64_t;
+    #define HAVE_I64
+  #endif
+  #if !defined _INTPTR_T_DEFINED
+    #if defined _LP64 || defined WIN64 || defined _WIN64
+      typedef __int64         intptr_t;
     #else
-      #include <stdint.h>
-    #endif
-  #elif !defined __STDC_VERSION__ || __STDC_VERSION__ < 199901L
-    /* The ISO C99 defines the int16_t and int_32t types. If the compiler got
-     * here, these types are probably undefined.
-     */
-    #if defined __MACH__
-      #include <ppc/types.h>
-      typedef unsigned short int  uint16_t;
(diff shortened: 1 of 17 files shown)


samphunter:
Wasn't there a problem with the PAWN 4 License? Or did you reimplement the feature?

Y-Less:
I didn't actually check that, since I did assume they were the same license.  However, I'll check the compatibility as they aren't the same.

Dayvison:
I like this, but i don't think zeex will merge
https://github.com/pawn-lang/compiler/pull/222#issuecomment-354535538

Y-Less:
@samphunter I don't think there's a problem.  I just copied several files wholesale, and they can be used unmodified in other projects.  This means that the VM will have different files with different licenses, but this is allowed and compatible.  The only thing is I need to modify amx.h a little bit more, that was the only one where parts were combined so I need to just use a modified version of one or the other, not a hybrid.

@Dayvison I think it would be nice to extend the scope, and don't see any harm in doing so, but that's not for me to decide.  @Zeex @Southclaws @YashasSamaga @Daniel-Cortez ?

samphunter:
@Y-Less there is already a repo that improves the runtime: https://github.com/Daniel-Cortez/pawn-3.2-plus

As for the license, shipping one application under two licenses is not the best, what would the compiled binary be licensed under? I am not a lawyer but that seems like a mess full of caveats. 

That being said, there are apparently already some features back-ported according to the what's new section.

Y-Less:
That project makes some backwards-incompatible changes, and doesn't have all the natives that this PR does.  I'd much rather keep these changes centralised if possible.

samphunter:
> And that statement from the 3.2 repo about licenses is just wrong, as long as the files aren't merged.

Yes, I do believe so as well. I wrote so in the PS. Though it may be worth considering, whether the Compiler Team could contact CompuPhase about getting the 3.2 code under the Apache License. I don't think they would particularly mind and having a more standard License may be beneficial.

Daniel-Cortez:
I think it's only logical to extend the project scope to the runtime improvement too. I mean, the group is already called 'pawn-lang', which implies that its activity should be focused on all aspects of the Pawn language, not only the compiler part.

Regarding this PR, the current license (zlib) is fairly permissive and simple, and I don't think mixing in new code under a much more complicated and restrictive license (Apache License 2.0) is a good idea - it would effectively put the whole repo under that new license (with some code still under the zlib license) since it's more restrictive than the current one and the end users would have to deal with its terms in the first place.

Also, @Y-Less, why did you need to backport the `amx_Address` macro? It's only used for physical addresses stored into the stack by `PUSHR*` opcodes, which aren't implemented in Pawn 3.2's instruction set.

> In the Readme of the pawn-3.2-plus repo, it is stated that back-porting is not possible due to the license difference.

I probably misphrased it back when I was writing that readme (I'm not a native English speaker). Technically mixing in code under the Apache License 2.0 should be possible if it's in separate files (but IMHO it's still not a good thing to do though).

> That project makes some backwards-incompatible changes

What changes do you mean exactly?

> and doesn't have all the natives that this PR does

The only functions from 4.0 my fork is missing are `fcopy`, `fcreatedir` and a few other functions that operate on *.ini files, all of them are from `amxfile.c`. There's also `floatint` in `amxfloat.c`, but it's not even referenced in `float.inc`. These functions shouldn't be too hard to reimplement without mixing in any code from Pawn 4.0. I might actually even add some of them tomorrow.

Zeex:
1. I'm a little confused about the purpose of this - are you embedding this runtime into somewhere else (i.e. not SA-MP)?

2. Do these natives work? I remember there was a breaking change in 4.0 in how AMX addresses are translated to `cell *` pointers (i.e. see `amx_Address`)

Y-Less:
>   I'm a little confused about the purpose of this - are you embedding this runtime into somewhere else (i.e. not SA-MP)?

Short answer, yes.

>   Do these natives work? I remember there was a breaking change in 4.0 in how AMX addresses are translated to cell * pointers (i.e. see amx_Address)

> Also, @Y-Less, why did you need to backport the `amx_Address` macro? It's only used for physical addresses stored into the stack by `PUSHR*` opcodes, which aren't implemented in Pawn 3.2's instruction set, so some of the natives you updated won't even work.

To answer you both at once, I included `amx_Address` because it was used by the new versions of some of the natives, but I edited it to work with DAT-relative offsets.  Those natives were just copy-pasted wholesale.

> What changes do you mean exactly?

Mostly the ones that restrict `LREF`/`SREF`, which would break YSI.  I know you can disable the new VM, but then there's not much point using that fork at all.

samphunter:
Well, as long as the compiler does not use any changes in the abstract machine and the machine is backward compatible, I guess there is no harm pulling the changes. 

As for me personally, I was playing around with implementing the plus runtime for samp, but I think the plus version is a much better fit. I am not sure for what the LREF/SREF hacking is needed, but I think it would be better to sacrifice it in order to properly sandbox scripts. Most of the YSI functionality can be provided by the native code, and can be implemented better in something like C++.

Y-Less:
Technically you could implement the YSI features in the compiler.  That is something I've thought about a few times, but just never got around to it.

samphunter:
> Technically you could implement the YSI features in the compiler. That is something I've thought about a few times, but just never got around to it.

Which features do you think would be appropriate to implement into the compiler? I can only think of inlines and hooks.

IMHO inlines are too big of a change to implement into this compiler, but hooks could be useful.

Y-Less:
inlines, hooks, and `___`.  They are the ones that use the most code generation to work.

Y-Less:
It would be a compiler fork though, or at least hidden behind a flag.

Daniel-Cortez:
> Mostly the ones that restrict LREF/SREF, which would break YSI.

Those are vulnerabilities, so I fixed them, as any other bugs.


> I know you can disable the new VM,

Or I can just make the abovementioned fixes optional. I was already asked to do so earlier and made some changes for it locally (added an `AMX_HARDENED` macro for conditional compilation of some of the security checks) but haven't tested these changes yet - I'm going to test and upload them later after I finish porting/reimplementing missing native functions from Pawn 4.0.

Speaking of function porting, in my fork I already copied `floatint` directly from Pawn 4 (CompuPhase doesn't hold the copyright for `amxfloat.c` so the license wasn't changed in this file) and added an overload for the `=` operator for automatic conversion from float to integer type (which wasn't done in 4.0 for some reason).
I also added `fcopy`, but implemented it in a different way - my implementation is not as simple as the original one in 4.0, but at least it's not vulnerable against a code injection attack (the original one calls `system("cp %s %s")` without escaping the file names).

Issue #296, opened 2018-03-25
refactoring of the codebase


YashasSamaga:
### Problems:
- constants are hardcoded
- poor coding practices (countless define macros)
- hacky solutions (checking for `static` qualifier using `fnumber`)

### TODO:
1. `enum` instead of defines for:
    ```
    #define taMULT      256 /* *= */
    #define taDIV       257 /* /= */
    #define taMOD       258 /* %= */
    #define taADD       259 /* += */
    #define taSUB       260 /* -= */
    .
    .
    .
    ```

    ```
    enum {
         taMULT = 256, /* *= */
         taDIV  =     257, /* /= */
         taMOD    =   258, /* %= */
         taADD    =   259, /* += */
         taSUB    =   260 /* -= */
    };
    ```
2. new scope mechanism
   We could use a new concept of *scope number* instead the  file number.  This would be useful for implementing the `#section` directive (and probably new features). It also makes the code legible and easier to maintain.

3. TBA

YashasSamaga:
In `chk_grow_litq`, the compiler does the following to resize the literal queue:

`litmax +=sDEF_LITMAX;`

This causes O(n^2) cells to be copied but this method is more space efficient in the worst case: max wastage = `sDEF_LITMAX - 1`.

The alternative solution is to:

`litmax *= 2;`

This causes O(n) cells to be copied but is less space efficient: half of the literal queue will be wasted.

**Tests:**
```
static arr[10000000] = {1, 2, 3, ...};
main () {
	arr[0] = 123;
}
```

`litmax+=sDEF_LITMAX;` takes on average around 13s (compiled continuously)
`litmax+=sDEF_LITMAX;` takes on average around 1.5s (compiled continuously)

YashasSamaga:
```
#define flgDEPRECATED 0x01  /* symbol is deprecated (avoid use) */
#define flagNAKED     0x10  /* function is naked */
#define flagPREDEF    0x20  /* symbol is pre-defined; successor of uPREDEF */
```

The `flags` member of `s_symbol` use just the above 3 bits. Is there a reason why they aren't continuous? (why isn't there `0x02`, `0x04`, etc.)

Daniel-Cortez:
>#define flgDEPRECATED 0x01  /* symbol is deprecated (avoid use) */
#define flagNAKED     0x10  /\* function is naked \*/
#define flagPREDEF    0x20  /\* symbol is pre-defined; successor of uPREDEF \*/
[...]

PR #300, opened 2018-05-07
compile-time assertions

### LINKED ISSUE #294: static assertions

* [ ] Bug Report
* [ ] Feature Request
* [ ] Question

**Design concept:**
`static_assert` takes a constant expression as the first parameter which must evaluate to true (non-zero number) for successful compilation and an *optional* error string as the second parameter.

`static_assert(cond, errstr="")`

The compiler would trigger a fatal error if `cond` evaluates to false. If an error string was given, the error string would be shown to the user; otherwise, the expression will be shown.

The compiler should allow `static_assert` to be embedded inside other constant expressions. For example,

`new arr[(static_assert(sz % 32 == 0, "array size must be a multiple of 32"), sz)];`
`const expr = 10 + !static_assert(size > 0);`

This can be achieved by making static_assert always return a value (say `1`). 

**Error messages:**

`test.pwn (123): fatal error 123: assertion failed: [expression]`
`test.pwn (123): fatal error 123: assertion failed: array size must be a multiple of 32`

**Why is it needed?**

This allows libraries to place assertions in the user's code (where `#error` isn't of much use). This also allows descriptive error messages.

YashasSamaga:
`static_assert(cond [, usermsg])`
`static_check(cond [, usermsg])`

`static_assert` triggers an error when `cond` evaluates to false

`static_check` triggers a warning when `cond` evaluates to false

Fixes #294 

Does a small bit of refactoring #296 by moving operator/keyword/etc. id numbers to an `enum`.

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

1.  **Test code 1:**
```
enum TEST {
	ABC1,
	ABC2
}
static_check(_:TEST == 2, "size of \"TEST\" is not 2");
static_check(_:TEST == 123, "size of \"TEST\" is not 123");

new Test[(static_check(1), TEST)];
static_check(sizeof(Test) == 2, "size of \"Test\" array is not 2");
static_check(sizeof(Test) == 123, "size of \"Test\" array is not 123");

main() {
	const var = 1023;
	new var2 = 1234;
	static_check(0);
	static_check(0, "test_string");
	static_check(0, "\\ test \\ \" abc \"");
	static_check(var%512 - 2*256 + 1, "this is zero");
	static_check(var - 1024, "this is not zero");
	static_check(var ^ var + 10 - 2*4 - 2);
	
	new arr[(static_check(32%2 == 0, "array size must be even"), 32)];
	new arr2[(static_check(33%2 == 0, "array size must be even"), 33)];
	
	static_check(var2);
	static_check(var, asd");
	static_check(var, "asd);
}
```

**Output (irrelevant messages have been removed):**
```
warning 239: check failed: size of "TEST" is not 123
warning 239: check failed: size of "Test" array is not 123
warning 239: check failed: 0
warning 239: check failed: test_string
warning 239: check failed: \ test \ " abc "
warning 239: check failed: this is zero
warning 239: check failed: var ^ var + 10 - 2*4 - 2
warning 239: check failed: array size must be even

error 008: must be a constant expression; assumed zero
error 001: expected token: "-string-", but found "-unknown-"
error 037: invalid string (possibly non-terminated string)
```

2. **Test code 2:**

```
static_assert(100 == 50, "100 is not equal to 50");
main() { }
```

**Output:**
`fatal error 110: assertion failed: 100 is not equal to 50`


AGraber:
Impatient for this feature

Y-Less:
I'm not sure about the syntax.  New compiler features should really use a `__` prefix to avoid possible clashes with user code - the double underscore is reserved for that reason.  Hence things like `__Pawn` and `__PanwBuild`.  Speaking of, it should also match the naming scheme of other compiler symbols, so `__StaticAssert` and `__StaticCheck` (I know there are other compiler symbols that don't follow this convention, but that's no reason to continue not doing so).

Y-Less:
Or we need to figure out which is preferable and maybe add `__pawn` and `__pawn_build`.

Daniel-Cortez:
>  New compiler features should really use a __ prefix to avoid possible clashes with user code

@Y-Less Not sure what you mean by "clashes" here.
Consider this code
```Pawn
#define static_assert(%0) \
	enum { __static_assert_ = 1/_:(%0) }

const SOME_CONST = 100;
static_assert(SOME_CONST > 0);
```

If there's a `static_assert` operator, it would just get shadowed by the macro, there would be no compile-time errors or anything.
The only thing the user would probably have to do is add an optional support for the new assertion operator and use the old `enum` trick as a fallback.
```Pawn
#define static_assert(%0,%1) \
	enum { __static_assert_ = 1/_:(%0) }
#if defined __PawnBuild
	#if __Pawn > 3 || __Pawn == 3 && __PawnBuild >= 9
		#undef static_assert
	#endif
#endif
```

And if the assertion operator gets renamed to `__static_assert`, it would actually make the code above slightly more compilcated
```Pawn
#define static_assert(%0,%1) \
	enum { __static_assert_ = 1/_:(%0) }
#if defined __PawnBuild
	#if __Pawn > 3 || __Pawn == 3 && __PawnBuild >= 9
		#undef static_assert
		#define static_assert __static_assert_ // <<== the user would also have to add this line
	#endif
#endif
```

Y-Less:
I never mentioned macros. Any variable or function in user code could be called that. All that can be guaranteed about user code is that they should never have a variable starting with `__`, because that is reserved for naming things in the compiler with no chance of ever conflicting.

Daniel-Cortez:
> Any variable or function in user code could be called that.

If a function is called `static_assert`, how is it supposed to be **static**_assert if it's a function? Sounds more like user's fault for naming a function in such misleading way. And pretty much the same logic with variables and constants.
While it's possible for any symbol to be named `static_assert`, in reality such situation should be highly unlikely for an adequate user. Adding `__` to the operator name (and thus compilcating the use of the operator by the need to type those extra two underscores every time) just to make sure that hipothetic situation never happens is like putting training wheels on a sport bike.

Y-Less:
There are multiple meanings of the word static. You can't just assume "I can't think of a reason to call something this, therefore no-one can". This is why naming conventions exist, so you don't have to.

Any new symbols added to the compiler should use the compiler naming convention. Otherwise: https://xkcd.com/1172/

And if you aren't using an editor with auto complete, then more fool you.

Y-Less:
In fact, "static" here is a whole new meaning of the word. There are three meanings I can think of:

static - restricted to a single scope
static - not dynamic, for things like hard-coded objects
static - immobile

So you can't claim that no-one could use the word static when it already has three common meanings in code.

Y-Less:
Maybe they just mis spelt asset... The point is, you can't know.

AGraber:
I don't think nobody cared of `static_assert` when it was added to C++11, and even if they did, it was still made part of it. If they misspelled asset, they no have the opportunity to correct it. I think if you're using this compiler you will want to get know all the new stuff it has (eventually you must turn on compatibility mode if you use old stuff), and not just use it because someone on the forum said "oh faster compile time!".

Y-Less:
> I don't think nobody cared of static_assert when it was added to C++11, and even if they did, it was still made part of it. If they misspelled asset, they now have the opportunity to correct it. I think if you're using this compiler you will want to get know all the new stuff it has (eventually you must turn on compatibility mode if you have some include that includes itself multiple times or something - you get the point), and not just use it because someone on the forum said "oh faster compile time!".

I think you've managed to seriously over-estimate EVERYONE there!

> I don't think nobody cared of static_assert when it was added to C++11

I'm sure they did.

> I think if you're using this compiler you will want to get know all the new stuff it has

I'm sure they won't.

> not just use it because someone on the forum said "oh faster compile time!".

Have you met the people on the forums?  If you told them a green font was faster they'd use it!

Daniel-Cortez:
>I was wondering if there was a neat replacement macro to handle this kind of use case:
new arr[static_assert(sz%2==0,"size needs to be even"),sz];

Well, the only other operator that can be used in expressions, statements and also outside of functions is `emit`, so...
```
#define static_assert(%0,%1) \
	emit const.pri (1 / _:(%0))
```
but it still wouldn't work with the standard compiler, so if you need compatibility you'll probably want to stick with the `enum`-based macro I posted above and not chain it within expressions.

YashasSamaga:
`#define static_assert __StaticAssert` or `#define static_assert __static_assert` would allow users to use the syntax they want but then there won't be a common syntax across many projects.

There is also the issue of consistency. We could break the consistency and introduce a new naming scheme. In future, if things like `addressof` or `nameof` or `attribute` are going to be added, they'll look like `__Addressof`, `__Nameof` and `__Attribute` respectively. The discussions in this PR is also applicable to the issues of the aforementioned feature requests.

Daniel-Cortez:
> `__Addressof`, `__Nameof` and `__Attribute`

What's the point in suddenly using CamelCase naming scheme, while currently all operators and built-in constants are named in lowercase (except for `__Pawn` and `__PawnBuild`, but those two are easier to ignore IMO, as it's probably too late to rename them to `__pawn` and `__pawnbuild` respectively)?


> 1. is this feature needed?

I don't see any reason why not. At the very least, it would be very useful in macros, where `#assert` can't be used.


> 2. what keyword has to be used

I'd prefer `static_assert`, but if compatibility (a.k.a. what-if-someone-names-a-variable-or-function-`static_assert`) is preferred more than usability, we could also provide `__static_assert` as a fallback for the compatibility mode (`-Z`).
Something like this:
  * With `-Z`: `__static_assert`.
  * Without `-Z`: `static_assert`, `__static_assert`.

Y-Less:
Compatibility should be concern number one for new compiler features!  The whole reason `-Z` exists is because that was ignored.

YashasSamaga:
`__Pawn` and `__PawnBuild` are compile-time constants but `static_assert` is an operator. Since other newly introduced operators such as `__emit` are not in Pascal case, it might be fine to use `static_assert` or `__static_assert`. (but then there is `emit` so...?)

Southclaws:
I think the important question to ask here is who *actually* cares whether some things are upper case and some things are lowercase? This is SA:MP not GCC - a few thousand people use it, not a few hundred thousand. Compatibility is key because the end users of this project probably care more about stability than whether symbols use a consistent case.

Y-Less:
I'm not too bothered if the final decision is __StaticAssert or __static_assert, as long as it has been discussed and consistent consensus reached. But consistency going forward and compatibility going back are must haves.

Y-Less:
You raise an interesting point about operators vs constants though. In normal code those should have naming schemes to differentiate them, maybe the same idea should be applied here.

Y-Less:
Alternatively, if you did want a flag, it should be the other way around - backwards compatible by default, then with `-b` (say) for "breaking changes" (which should really have been how `-Z` was done).

Daniel-Cortez:
> You're talking like typing __ is some huge effort that will reduce coding 500%.

I never said it would be very hard to type `__`, but I still find it excessive to complicate the use of the operator just because of some very unlikely hypothetical situation.

> I think the important question to ask here is who actually cares whether some things are upper case and some things are lowercase?

The end users, probably? Because if we start naming new operators in CamelCase, the users would have to keep in mind that while most of the operators are in lower case, some of the new ones are named in CamelCase for some weird reason, which wouldn't be very user-friendly.

AGraber:
lowercase with underscore for word breaks seems perfect (since we have __emit anyways). The `__` dilema can be avoided with a simple define after all.

One question though, since there's already something for a user-defined `emit` to take precedence over the compiler's, would it be feasable to repeat that here too?

YashasSamaga:
Do we go with `__static_assert` and `__static_check`?

Y-Less:
I'd say yes.

Y-Less:
Or ________sTaTiC_ASSert

source/compiler/sc2.c
```diff
@@ -2090,7 +2090,7 @@ char *sc_tokens[] = {
          "assert", "*begin", "break", "case", "char", "const", "continue", "default",
          "defined", "do", "else", "emit", "__emit", "*end", "enum", "exit", "for",
          "forward", "goto", "if", "native", "new", "operator", "public", "return",
-         "sizeof", "sleep", "state", "static", "static_assert", "static_check", "stock",
+         "sizeof", "sleep", "state", "__static", "__static_assert", "static_check", "stock",
```
Daniel-Cortez: @YashasSamaga Probably not what you really intended...

YashasSamaga: Yup, realized it while writing the test. Thanks.

YashasSamaga: What's worse is that the `WILL_FAIL` testing method does not work with regular expression tests (which means pretty much all the merged tests for errors are wrong)

source/compiler/sc2.c
```diff
@@ -2141,8 +2141,8 @@ char *sc_tokens[] = {
          "assert", "*begin", "break", "case", "char", "const", "continue", "default",
          "defined", "do", "else", "emit", "__emit", "*end", "enum", "exit", "for",
          "forward", "goto", "if", "native", "new", "operator", "public", "return",
-         "sizeof", "sleep", "state", "static", "stock", "switch", "tagof", "*then",
-         "while",
+         "sizeof", "sleep", "state", "__static", "__static_assert", "static_check", "stock",
```
Zeex: What is `__static`?

source/compiler/sc3.c
```diff
@@ -1390,6 +1390,14 @@ static int hier2(value *lval)
     while (paranthese--)
       needtoken(')');
     return FALSE;
+  case t__STATIC_ASSERT:    
+    static_check(FALSE);
+    pc_sideeffect=TRUE;
+    return FALSE;
+  case t__STATIC_CHECK:
```
Zeex: Since`static_check` has no underscores, why is its token called `t__STATIC_CHECK`?

source/compiler/sc5.c
```diff
@@ -422,3 +424,67 @@ int pc_geterrorwarnings()
...
 }
 
+/* static_check()
+ * checks compile-time assertions and triggers an error or warning
+ *
+ * Parameters:
+ * FALSE for error if assertion fails
+ * TRUE for warning if assertion fails
+ */
+int static_check(int warning)
```
Zeex: Is sc5.c a good place for this function? I don't see any other parsing business in this file, it looks like it is intended to hold the error reporting system, parsing should be done elsewhere. 

sc1.c might be a better place, we already have `doassert` in there. Also, how about renaming this function to `do_static_assert`? So that it's more consistent with other functions.

Zeex:
Can `static_assert` be used without parenthesis, similar to `assert`?

```pawn
main() {
	new x = -1;
	assert x > 0;
}
```

source/compiler/sc2.c
Y-Less: It looks like the `__` was added on the wrong symbol, it should be:

```c
"static", "__static_assert", "__static_check"
```

YashasSamaga: That was a typo while adding the underscore prefixes.

source/compiler/sc3.c
YashasSamaga: That was a typo while adding the underscore prefixes.

YashasSamaga:
1. Will make the requested changes.

2. Will make the parenthesis optional.

One serious issue with the testing system is that we can't test for errors as the compiler returns a non-zero exit code. This causes CTest to mark the test as failed even though the non-zero exit code was intended.

By setting `WILL_FAIL` property to `TRUE`, CTest won't consider non-zero exit code as a failure (but a crash is still a failure). However, this has a side effect that the test is considered passed even if the regular expression test fails.

Googling the issue up leads to answers which suggest writing a wrapper around the compiler to deal with this kind of issues. I am not sure how we should go about it.

Some of the tests that were added from the recently merged PRs allow the test to fail (test script and expected output are present but failures are allowed to deal with non-zero exit code). Once a solution is found these tests must be updated.

YashasSamaga:
Since the user message string is optional, an ambiguity is created: `__static_check 0, "a";`

Should it be treated as `__static_check(0), "a"` or as `__static_check(0, "a")`

Southclaws:
At the risk of shilling my own code (sorry!) - perhaps the testing could be enhanced with sampctl?

YSI already provides a testing package and sampctl is already set up to parse the y_testing output and produce exit codes depending on whether the tests passed or failed - I'm just thinking it might be simpler to re-use an existing tool than try to roll one with CMake or bash.

Links:
- https://github.com/Southclaws/sampctl/wiki/Runtime-Modes#y_testing
- https://travis-ci.org/Southclaws/samp-ini

Maybe worth opening an issue for? cc @YashasSamaga

YashasSamaga:
The latest commit turns on the staging buffer if it wasn't already on before evaluating (inside `constexpr()` function) a constant expression. Previously, the staging buffer would use to be always enabled on entering `constexpr()` and then disabled before returning. This would fail to work if the staging buffer was already enabled preventing few good use cases for `static_assert/static_check`.

I am not sure why `constexpr()` originally wasn't considering the possibility of the staging buffer already turned on.

This fix allows the following use-case:

```
#define nameof(%0) (__static_check(defined(%0)),#%0)

forward mypublic();
public mypublic() { }

native SetTimer(const fname[], interval, repeat = false);
native printf(const format[], ...);

main () {
	SetTimer(nameof(mypublic), 1000, false);
}
```

YashasSamaga:
## DO NOT MERGE

Working on allowing string concatenation for the user string.

source/compiler/sc.h
```diff
@@ -311,108 +311,119 @@ typedef struct s_valuepair {
...
+  tLAST = tpWARNING, /* value of last multi-character match-able token */
+
+  /* semicolon is a special case, because it can be optional */
+  tTERM,/* semicolon or newline */
+  tENDEXPR, /* forced end of expression */
+  /* other recognized tokens */
+  tNUMBER,/* integer number */
+  tRATIONAL, /* rational number */
+  tSYMBOL,
+  tLABEL  ,
```
Daniel-Cortez: Any reason to have two whitespaces between `tLABEL` and a comma?

source/compiler/sc1.c
```diff
@@ -5490,15 +5498,18 @@ static int doexpr(int comma,int chkeffect,int allowarray,int mark_endexpr,
  */
 SC_FUNC int constexpr(cell *val,int *tag,symbol **symptr)
 {
-  int ident,index;
+  int ident,index,already_staging;
   cell cidx;
 
-  stgset(TRUE);         /* start stage-buffering */
+  already_staging=stgget(&index,&cidx);
+  if(!already_staging)
```
Daniel-Cortez: No whitespace between `if` and a `(`.

source/compiler/sc1.c
```diff
@@ -5998,6 +6009,75 @@ static void dolabel(void)
...
+ *
+ * Parameters:
+ * FALSE for error if assertion fails
+ * TRUE for warning if assertion fails
+ */
+SC_FUNC int do_static_check(int warning)
+{
+  cell val;
+  int noerrors=TRUE,usermsg_present=FALSE,usermsg_len=0;
+  char usermsg[sLINEMAX+1];
```
Daniel-Cortez: Is it possible to use `__static_assert` within another `__static_assert`, like `__static_assert(__static_assert(1), 1)`? And if so, wouldn't allocating an array of size `sLINEMAX+1` for two or more times lead to a stack overflow?

YashasSamaga: For the record, `__static_check` and `__static_assert` return the value of the constant expression that was evaluated, i.e. `new var = __static_check(expr);` is equivalent to `new var = expr;` except for the potential warning that could be triggered.

Daniel-Cortez: > Stack sizes for user-mode threads are generally much more than 8KB.

This is probably true for the major OSes such as Windows, Linux and macOS, but the compiler can potentially target many more platforms (since it only uses the standard C library and POSIX functions on unknown platforms), so we can never know if allocating 4096 bytes (actually a little bit more since there are also other local variables within that function) from the stack for more than once is safe.

Maybe it would be better to use `malloc` instead of a static array? I personally see no harm from it since there might even be no need to allocate anything if the user didn't specify an error message. There's also `alloca`, but it's not very portable and it's only used within some of the AMX extension modules; `amxCore`, `amxFixed`, `amxFloat` and `amxTime` modules and the interpreter itself don't seem to rely on it.

Y-Less: If you are using the compiler on a more constrained system, I'd think you'd want to drop `sLINEMAX` way down in the first place.

Daniel-Cortez: Which 1) would require interference with modifications to the source code, and 2) still could be unnecessary if the stack size is the only limiting factor. Currently `do_static_check` is the only function that uses so much of stack space.

PR #303, opened 2018-05-11
directory default include

### LINKED ISSUE #295: directory defaults

* [ ] Bug Report
* [x] Feature Request
* [ ] Question

**What happened**:

The file `default.inc` is automatically included when you compile.  It would be nice to extend this to other directories - similar to the `index.js` feature in JS 6:

Current file structure in YSI:

```
- YSI_Coding
    - y_hooks
        - impl.inc
        - macros.inc
        - tests.inc
    - y_hooks.inc
- YSI_Data
    - y_foreach
        - impl.inc
        - iterators.inc
        - macros.inc
        - tests.inc
        - warning.inc
        - yield.inc
    - y_iterate.inc
```

If doing `#include <directory>` looked for a `default.inc` in that directory the code could change to:

```
- YSI_Coding
    - y_hooks
        - impl.inc
        - macros.inc
        - tests.inc
        - default.inc
- YSI_Data
    - y_iterate
        - impl.inc
        - iterators.inc
        - macros.inc
        - tests.inc
        - warning.inc
        - yield.inc
        - default.inc
```

Which would put less stuff in the root directory.

In terms of `-Z`, the guard should be generated as if the filename was purely the containing directory.

**Environment**:

* Operating System
* Compiler version

YashasSamaga:
**[DO NOT MERGE - WORK IN PROGRESS]**
--

This commit makes the compiler look for `default.[extension]` in the given include path if the search for include file fails.

For example, <test> makes the compiler first look for test.inc in include directories. If the search fails, it then looks for `test/default.[extension]`.

Fixes #295

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -331,6 +331,11 @@ static void doinclude(int silent)
      * between <...> are only read from the list of include directories.
      */
     result=plungefile(name,(c!='>'),TRUE);
+    if(!result && (c=='>')) {
+      strlcat(name,"/default",sizeof(name));
+      result=plungefile(name,(c!='>'),TRUE);
+      name[i]='\0';
+    }
     if (result && pc_compat)
       add_constant(symname,1,sGLOBAL,0);
     else if (!result && !silent)


YashasSamaga:
There's this bit of code right above:
```
 char dirsep=
      #if DIRSEP_CHAR!='\\'
        '\\';
      #else
        DIRSEP_CHAR;
      #endif
```
I think there's a typo there which makes `dirsep` to have `\\` in either case.

Since `/` works both in Windows & Linux, I decided to use `/` because I suspected using `dirsep` would fail on Linux. I am not sure what the aforementioned bit of code is supposed to do and how to fix the typo or whatever that caused the redundancy.

Y-Less:
Well no, the check should just be changed to ` TRUE `. And I'm just concerned about how always using `/` will affect the current directory and file name resolution as in the include guards wiki page.

YashasSamaga:
What is used as the directory separator for `default` file has no affect on the include guard generated.

If the user does `#include<ABC\EFG/HIJ/KLM>` and `KLM.inc` does not exist but `default.inc` exists in the `KLM` folder, the include guard generated is same as the one that would be generated if `KLM.inc` existed.

Y-Less:
Yes, but what about other files included from that one?

PR #311, opened 2018-06-08
destructor not implemented error

### LINKED ISSUE #310: Destructor without implementation targets incorrect function

* [x] Bug Report
* [ ] Feature Request
* [ ] Question

**What happened**:

So I was messing with destructors for my pawn-errors package, eventually didn't end up using them so I removed the implementation for `operator~(Error:e[], s)` however I forgot that I left the `forward operator~(Error:e, s);` in the source file, this resulted in some very odd behaviour.

I would have expected the same as if you omit any implementation, you get an error message saying "function f(x) is not implemented" however that doesn't seem to apply to `operator~`.

```pawn
#include <a_samp>


forward operator~(Error:right[], size);

stock Error:Error(a, what[]) {
    printf("Error called with %d '%s'", a, what);
    return Error:a;
}

main() {
    new Error:e = Error(1, "hi");
}
```

Side note: if you remove the assignment in `main` so it's just a declaration of `Error:e` then the server crashes once it reaches the end of `main()` with `Stack/heap collision: insufficient stack size`.

I've not tried this with much other code, I did try it without the `a` parameter in the above repro but it still crashed. It's very strange because the signatures do not match at all, not by tag and not by parameter types.

**Environment**:

* Operating System: Windows, not tested others
* Compiler version: 3.10.7
* How are you invoking the compiler? sampctl and command line, standard arguments

YashasSamaga:

```
#include <a_samp>

forward operator~(Error:right[], size);
main() {
    new Error:e;
}
```

```
CODE 0    ; 0
;program exit point
    halt 0

    proc    ; main
    ; line 4
    ; line 5
    break    ; c
    ;$lcl e fffffffc
    push.c 0
    ;$exp
    push.pri
    push.c 1
    addr.pri fffffffc
    push.pri
    push.c 8
    call .~4000000f    ; operator~(Error:)
    pop.pri
    stack 4
    zero.pri
    retn

STKSIZE 1000
```

The compiler tries to call a non-existent function. This commit instead throws "function not implemented" error when destructor definition is missing.

`error 004: function "operator~(Error:)" is not implemented`

Fixes #310 

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -4967,6 +4967,14 @@ static void destructsymbols(symbol *root,int level)
       /* check that the '~' operator is defined for this tag */
       operator_symname(symbolname,"~",sym->tag,0,1,0);
       if ((opsym=findglb(symbolname,sGLOBAL))!=NULL) {
+        if ((opsym->usage & uMISSING)!=0 || (opsym->usage & uPROTOTYPED)==0) {
+          char symname[2*sNAMEMAX+16];  /* allow space for user defined operators */
+          funcdisplayname(symname,opsym->name);
+          if ((opsym->usage & uMISSING)!=0)
+            error(4,symname);           /* function not defined */
+          if ((opsym->usage & uPROTOTYPED)==0)
+            error(71,symname);          /* operator must be declared before use */
+        } /* if */
         /* save PRI, in case of a return statment */
         if (!savepri) {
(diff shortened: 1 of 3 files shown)


YashasSamaga:
PR is ready to be merged.

Requires approval.

Southclaws (approved):

PR #312, opened 2018-06-09
fix result propagation in constexpr operator chains

### LINKED ISSUE #308: Conditional fail with constant numbers

* [x] Bug Report
* [ ] Feature Request
* [ ] Question

**What happened**:
The condition should give true, but instead give false

```pawn
new min1 = 15;
new max1 = 30;
#define min2  15
#define max2  30
main()
{
    printf("%d", (min1 < 20 < max1));
    printf("%d", (min2 < 20 < max2));
    printf("%d", (15 < 20 < 30));
}
```
Const, define and explict numbers are the same
`new` + `const` works
```pawn
new const min = 15;
new const max = 30;
main()
{
    printf("%d", (min < 20 < max));
}
```
**Environment**:

* Operating System Linux
* Compiler version 3.10.7(All versions do the same)
* How are you invoking the compiler? Pawno, Sublime, vscode, sampctl or command-line? command-line

YashasSamaga:

The compiler fails to propagate the result of chained operations through the entire chain.

`15 < 20 < 30` evaluates to false even though it is true.

This PR resolves the above issue.

Refer to the commit description and https://github.com/pawn-lang/compiler/issues/308#issuecomment-395941341 for details.

Fixes #308 

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -512,6 +512,7 @@ static int plnge_rel(int *opstr,int opoff,int (*hier)(value *lval),value *lval)
       error(212);
     if (count>0) {
       relop_prefix();
+      lval2.boolresult=lval->boolresult;
       *lval=lval2;      /* copy right hand expression of the previous iteration */
     } /* if */
     opidx+=opoff;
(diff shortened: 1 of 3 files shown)


YashasSamaga:
This PR is ready to be merged.

Hit LIKE if you approve; otherwise hit DISLIKE.

Southclaws (approved):

Southclaws:
I think appveyor got stuck (been running since this morning) so I just forced merge.

PR #334, opened 2018-07-01
Updated readme.md file and added forum thread bbcode

Southclaws:

This is from a discussion I had with some of the team a while ago regarding trying to push this project a bit more in the wider SA:MP community.

So far, there is virtually **no** forum presence of this project outside a very old thread and brief mentions of it in threads.

The team (not me! I just do the housekeeping, I'm worthless with compilers and C!) puts in a *ton* of really great work in this project and I feel that it's slightly wasted because only a subset of the wider community (mostly those who hang around on the Discord server) know about this project and its advantages.

This proposal isn't just for an updated readme, I propose we post the `readme.bbcode` onto the forums and actively maintain the thread with updates so we can get feedback from people who solely use the forums and don't step into the "advanced" side of Pawn (Discord, GitHub, etc).

I'd be willing to keep this thread updated. I rendered the bbcode using "forumfmt", a tool for converting Markdown to bbcode.

Another issue I wanted to bring up was the naming. Most of the time, people call this "Zeex compiler" or "new compiler" and I think it would be good if we settled on a proper name for this project that we can refer to it. Zeex obviously did some amazing work for years getting this project to the state it was at before it was taken on by a team. But I think it would be good if we re-phrase how we refer to it so people know it's now the work of multiple people, not just one.

I propose we refer to this project as "The Community Pawn Compiler" or simple "community compiler" in threads, issues and Discord. There is also a bit of a myth floating around the forums that "The Zeex Compiler" is notoriously difficult to install and buggy, I think "re-branding" it would partly resolve some of these myths and get people interested in it again.

* [ ] A Bug Fix
* [ ] A New Feature
* [x] Some repository meta (documentation, etc)
* [ ] Other

**Additional Documentation**:

I could add this to the readme if necessary:

To re-generate the bbcode, install [forumfmt](https://github.com/Southclaws/forumfmt) and run:

```bash
forumfmt readme.md readme.bbcode
```


BigETI:
Re-branding it under a new name with a better image over the entire SA:MP community is a good idea to be quite honest. Over the years this project has become superior over the old tool set we used to work with. I believe that this can be a standard tool for all SA:MP server developers, including PAWN library developers.

readme.md
```diff
@@ -1,58 +1,119 @@
...
+This is a modified copy of the Pawn compiler version 3.2.3664 by Compuphase that
+fixes some bugs and adds a few features.
+
+This project was originally maintained and managed by Zeex who left the project
+due to lack of time on 2017-12-31. Thank you for all your hard work on keeping
+this project alive Zeex!
+
+The project is now maintained by the
+[Compiler Team](https://github.com/orgs/pawn-lang/teams/compiler) at the
+pawn-lang GitHub org.
```
YashasSamaga: These frequent unnecessary line breaks are very ugly when viewed in an editor (even though markdown treats them to be part of a single line).

readme.md
```diff
@@ -1,58 +1,119 @@
...
+}
+```
 
 ## Building from source code
 
-In general you will need [CMake](https://cmake.org/) and a C compiller to build Pawn from source code.
+In general you will need [CMake](https://cmake.org/) and a C compiller to build
+Pawn from source code.
 
 ### Building on Windows
```
YashasSamaga: Visual Studio 2017 and later has CMake integrated into the IDE. The folder containing CMakeLists.txt has to be opened in Visual Studio. The IDE will automatically generate the cache and allow you to build in one click.

readme.md
```diff
@@ -1,58 +1,119 @@
...
 
-This is a modified copy of the Pawn compiler version 3.2.3664 by Compuphase that fixes some bugs and adds a few features.
+The reason this exists is to:
 
-This project was originally maintained and managed by Zeex who left the project due to lack of time on 2017-12-31. Thank you for all your hard work on keeping this project alive Zeex!
+- Fix known bugs with the original compiler
+- Provide a better development experience for the SA:MP community
 
-The project is now maintained by the [Compiler Team](https://github.com/orgs/pawn-lang/teams/compiler) here at the pawn-lang GitHub org.
+You shouldn't have any reason _not_ to use this compiler. If you find problem,
```
YashasSamaga: I think this actually bloats it with not-so-useful information. A link to the list of features is sufficient, IMO.

readme.md
```diff
@@ -1,58 +1,119 @@
...
+[Compiler Team](https://github.com/orgs/pawn-lang/teams/compiler) at the
+pawn-lang GitHub org.
 
 [Original readme](readme_compuphase.txt)
 
-## What is this?
+## Why
 
-This is a modified copy of the Pawn compiler version 3.2.3664 by Compuphase that fixes some bugs and adds a few features.
+The reason this exists is to:
```
YashasSamaga: This sounds very informal and a bit awkward (just my opinion). The original line could be rephrased as "This is a modified version of the Pawn 3.2.3664 compiler with many bug fixes and enhancements."

readme.md
```diff
@@ -1,58 +1,119 @@
...
+
+## What
+
+This is a modified copy of the Pawn compiler version 3.2.3664 by Compuphase that
+fixes some bugs and adds a few features.
+
+This project was originally maintained and managed by Zeex who left the project
+due to lack of time on 2017-12-31. Thank you for all your hard work on keeping
+this project alive Zeex!
+
```
YashasSamaga: I am not really sure if @Zeex has fully given up on this project. I think it would be sufficient if it states that Zeex was the founder of this project or something along those lines (just my opinion) with some bits of history.

readme.md
```diff
@@ -122,14 +189,21 @@ make
 
 ## Background
 
-The project was originally started as a set of patches aimed to create a compiler that would be compatible with the compiler used in [SA-MP (San Andreas Multiplayer)](http://sa-mp.com/).
+The project was originally started as a set of patches aimed to create a
```
YashasSamaga: Shouldn't this section be the first one? (instead of reasons why this exists)

readme.md
```diff
@@ -122,14 +189,21 @@ make
...
+bugs can't be fixed. So the main goal of the project is to re-create changes
+that were previously made by the devs as well as fix all known compiler bugs.
 
-[1] It's worth noting that the version of the AMX embedded into the SA-MP server seems to be based on an older release of Pawn.
+[1] It's worth noting that the version of the AMX embedded into the SA-MP server
+seems to be based on an older release of Pawn.
 
 [build]: https://travis-ci.org/pawn-lang/compiler
 [build_status]: https://travis-ci.org/pawn-lang/compiler.svg?branch=master
 [build_win]: https://ci.appveyor.com/project/Southclaws/compiler/branch/master
```
YashasSamaga: AppVeyor link needs to be changed.

YashasSamaga:
In my opinion, leaving the readme as it is with just the name changed and sampctl section added is better.

readme.md
Southclaws: This is markdown lint rules, wrapping at column 80. My editor does it automatically.

readme.md
Southclaws: Yes I will update this with that info. It's far easier than doing the generation beforehand!

readme.md
Southclaws: I think potential users will care more about a value proposition than the version number. I will see about rephrasing it though.

readme.md
Southclaws: Yes this is rather outdated since it was written when Zeex handed off the repo to me.

readme.md
Southclaws: Thanks for reminding me!

readme.md
Southclaws: I'll move all the existing links to references at the end of the document to clear this up.

readme.md
```diff
@@ -1,86 +1,146 @@
...
+
+- [What's new][new] contains features and other notable changes.
+
+- [Release notes][releases] list of all official releases of the compiler
+  binaries.
+
+## How to Use
 
-[Release notes](https://github.com/pawn-lang/compiler/releases) descibe changes in each version of the compiler.
+Binary packages can be downloads from [Releases][releases].
```
Zeex: Looks like a typo, "downloads" should be "downloaded"

Southclaws:
I realised one of the main reasons for doing this was to point out that people need to use `-Z+` so they don't have to dig through the wiki or ask on Discord/forum why their includes aren't working.

Southclaws:
If everyone is happy with the changes, I'll merge this and get a topic on the forum and keep it topped up with updates regarding the compiler.

Y-Less:
Do we even need the bb code file?  That's now trivial to generate from the markdown.

Southclaws:
Probably not, I just tend to commit them so I have a copy available. I'll remove it since your browser script does the job nicely anyway!

Y-Less:
Any objections to merging this?

Zeex:
Looks ok to me 👍 

I feel like the readme is getting big, maybe adding an index at the top could make it easier to read. But that's just some idea (not necessary to do in this PR).

Southclaws:
Now that you mention it, it is quite large.

I'm thinking of the type of people who will land on this page (or the thread, if everyone agrees on posting one). I don't think the majority of users are looking for instructions to build the compiler itself so maybe that whole section could be moved to a wiki page? I think the important things that should exist here is an answer to the repeated "why are my includes not working properly" which the obvious (to us) answer is `-Z+`. In short, a quick "getting started" would suffice I think.

Also, just a note to me: I need to elaborate the sampctl section and explain that sampctl already uses this compiler by default.

Zeex:
Yeah, build and installation stuff could be moved to the wiki or separate .md files (with links added to the main readme probably).

Zeex:
Regarding `-Z`...  if most people need to use this flag to compile their code, maybe its default value should be changed to "off"? Maybe it was a dumb idea after all. But I don't know how many people would be upset about this change. Adding a message like "compiling in compatibility mode, use -z to turn it off" could make this less painful for them (yes, small z).

Southclaws:
I moved build stuff to a wiki page, I removed some unnecessary sampctl stuff (sampctl users already use this compiler so I don't really know why I added that section...) and I fixed some other minor issues.

I realised there's already a paragraph regarding `-Z` in the Windows "How to Use" section - is `-Z` Windows-specific?

Y-Less:
I think `-Z` is just here to stay now.  While future breaking changes should be behind compiler flags, most major code has adapted and most people know about it now.  Changing it again would IMHO be worse than leaving it.

Southclaws:
If we made `-Z` `+` by default, I don't think there would be any negative consequences. People who already have `-Z+` in their args would be fine and people who already just have `-Z` would be fine. Anyone who for some reason doesn't need it (they don't use sscanf, YSI, or any libraries that have trouble) probably won't be affected either because those users probably aren't going to have anything outside of the standard `pawno/include/` structure with mostly default libraries.

And on the plus side, new users won't get confused about why they are seeing errors that seemingly make no sense and likely just switch back to the old one without bothering to figure it out.

Southclaws:
Okay, I'm pretty happy with it now, if you guys are I'll merge to master and get a thread up (if the forum ever comes back online...)

PR #335, opened 2018-07-02
Don't generate function preamble on naked functions

### LINKED ISSUE #324: Skip `PROC` with `#pragma naked`

* [ ] Bug Report
* [x] Feature Request
* [ ] Question

**What happened**:

Functions are generated with `PROC` as the first instruction.

Not to be - make the completely naked, no preamble or postamble.

**Environment**:

* Operating System
* Compiler version

AGraber:

- Skips function preamble (`PROC` instruction) on naked functions

Fixes #324 

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:
Assembly output demonstrating the new feature:

test1.pwn:
```pawn
#pragma naked
f(a)
{
    #emit LOAD.S.pri a
    #emit RETN
}

main()
{
	f(5);
}
```

test1.asm:
```asm

CODE 0	; 0
;program exit point
	halt 0

	; f
	; line 3
	;$lcl a c
	load.s.pri c
	retn 

	proc	; main
	; line 9
	; line a
	break	; 18
	push.c 5
	;$par
	push.c 4
	call .f
	;$exp
	zero.pri
	retn


STKSIZE 1000
```

test2.pwn:
```pawn
f(a)
{
    #emit LOAD.S.pri a
    #emit RETN
}

main()
{
	f(5);
}
```

test2.asm:
```asm

CODE 0	; 0
;program exit point
	halt 0

	proc	; f
	; line 2
	;$lcl a c
	load.s.pri c
	retn 
	zero.pri
	retn

	proc	; main
	; line 8
	; line 9
	break	; 24
	push.c 5
	;$par
	push.c 4
	call .f
	;$exp
	zero.pri
	retn


STKSIZE 1000
```

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -640,7 +640,7 @@ SC_FUNC void setlinedirect(int line);
 SC_FUNC void setlineconst(int line);
 SC_FUNC void setlabel(int index);
 SC_FUNC void markexpr(optmark type,const char *name,cell offset);
-SC_FUNC void startfunc(char *fname);
+SC_FUNC void startfunc(char *fname,int generateproc);
 SC_FUNC void endfunc(void);
 SC_FUNC void alignframe(int numbytes);
 SC_FUNC void rvalue(value *lval);
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -3798,7 +3798,7 @@ static int newfunc(char *firstname,int firsttag,int fpublic,int fstatic,int stoc
       ptr=ptr->next;
     } /* while */
   } /* if */
-  startfunc(sym->name); /* creates stack frame */
+  startfunc(sym->name,(sym->flags & flagNAKED)==0); /* creates stack frame */
   insert_dbgline(funcline);
   setline(FALSE);
   if (sc_alignnext) {
--- a/source/compiler/sc4.c
+++ b/source/compiler/sc4.c
@@ -343,9 +343,12 @@ SC_FUNC void markexpr(optmark type,const char *name,cell offset)
  *
  *  Global references: funcstatus  (referred to only)
  */
-SC_FUNC void startfunc(char *fname)
+SC_FUNC void startfunc(char *fname,int generateproc)
 {
-  stgwrite("\tproc");
+  if (generateproc) {
+    stgwrite("\tproc");
+    code_idx+=opcodes(1);
+  } /* if */
   if (sc_asmfile) {
     char symname[2*sNAMEMAX+16];
     funcdisplayname(symname,fname);
@@ -353,7 +356,6 @@ SC_FUNC void startfunc(char *fname)
     stgwrite(symname);
   } /* if */
   stgwrite("\n");
-  code_idx+=opcodes(1);
 }
 
 /*  endfunc


YashasSamaga:
It would be useful if the function name was printed. The `startfunc` prints the function name along with the `PROC` instruction.

Y-Less:
Needs tests.

YashasSamaga:
Instead of adding code to directly write thr function name to the staging buffer, you could have passed an argument to `startfunc` to inform if `PROC` (or only name) had to be printed. It won't be difficult to change since `startfunc` is called at just one or two places.

This (IMO) gives more readable code and also prevents low level code generation functions polluting the parser code.

AGraber:
I didn't knew if doing that (adding an argument) would be correct in this project (given the fact that those functions haven't been touched in years!), but now I know it's OK. I'll edit the branch according to that then, and I'll also take a look at how tests are made in order to add them soon.

YashasSamaga:
I think the current testing setup does not allow these kinds of issues to be tested. This requires reading the contents of a file. 

The following test scenarios exist:
1. testing compiler output
2. testing compiler exit status 

The following test scenarios don't exist:
3. testing assembly generated
4. execute the script and collect results (can be achieved with 3. but would be insane to test more complex code that way)

Type 3 is what this PR needs.

cc @Y_Less @Southclaws @Zeex

Southclaws:
Yes I agree, I also have no idea how to add tests (I'm also not a CMake expert). The process should be documented on the wiki or contributing guide.

AGraber:
I remade the branch to avoid trash commits and came up with a more cleaner approach that doesn't touch `sc1.c` too much, while still printing the function name when functions are naked. Also updated the PR message to reflect the assemblies that are currently generated by this build.

Y-Less:
Unfortunately you're right about the difficulty in testing this in the current setup - and fixing that is well outside the scope of this feature.  So for now I'd say we might have to skip the testing of this, bad as that may be.

source/compiler/sc.h
```diff
@@ -640,7 +640,7 @@ SC_FUNC void setlinedirect(int line);
 SC_FUNC void setlineconst(int line);
 SC_FUNC void setlabel(int index);
 SC_FUNC void markexpr(optmark type,const char *name,cell offset);
-SC_FUNC void startfunc(char *fname);
+SC_FUNC void startfunc(char *fname, char isnaked);
```
Daniel-Cortez: The compiler codebase usually utilizes `int` for boolean variables, not `char`.

YashasSamaga: The styling used in the project requires that there isn't a space between arguments.

source/compiler/sc1.c
```diff
@@ -3798,7 +3798,7 @@ static int newfunc(char *firstname,int firsttag,int fpublic,int fstatic,int stoc
       ptr=ptr->next;
     } /* while */
   } /* if */
-  startfunc(sym->name); /* creates stack frame */
+  startfunc(sym->name, sym->flags & flagNAKED); /* creates stack frame */
```
YashasSamaga: The styling followed by the project requires that you do not add a space between arguments.

source/compiler/sc4.c
```diff
@@ -343,17 +343,19 @@ SC_FUNC void markexpr(optmark type,const char *name,cell offset)
  *
  *  Global references: funcstatus  (referred to only)
  */
-SC_FUNC void startfunc(char *fname)
+SC_FUNC void startfunc(char *fname, char isnaked)
```
YashasSamaga: The styling followed by the project requires that you do not add a space between arguments.

source/compiler/sc4.c
```diff
@@ -343,17 +343,19 @@ SC_FUNC void markexpr(optmark type,const char *name,cell offset)
...
 {
-  stgwrite("\tproc");
+  if(!isnaked) {
+    stgwrite("\tproc");
+    code_idx+=opcodes(1);
+  }
```
YashasSamaga: The styling followed by the project requires that you add a `/* if */` at the closing brace.

source/compiler/sc.h
YashasSamaga: I am being evil but I think the parameter makes more sense if it is related to the prologue than nakedness of a function?

source/compiler/sc4.c
```diff
@@ -343,17 +343,19 @@ SC_FUNC void markexpr(optmark type,const char *name,cell offset)
  *
  *  Global references: funcstatus  (referred to only)
  */
-SC_FUNC void startfunc(char *fname)
+SC_FUNC void startfunc(char *fname,int generateproc)
 {
-  stgwrite("\tproc");
+  if(generateproc) {
```
Daniel-Cortez: The coding style followed by the project requires a space between `if`, `while`, `switch` etc. and an opening parenhesis (e.g. `if (expr)`, `while (expr)`, `switch (expr)`).

YashasSamaga:
```
f() { }

#pragma naked
g() { }

main () {
  f();
  g();
}
```

produces

```
proc	; f
	zero.pri
	retn

	; g
	; line 4

	proc	;
	push.c 0
	call .f
	push.c 0
	call .g
	zero.pri
	retn
```

The programmer can be blamed for that I guess because the function was marked as naked but no return statement was added.

This PR is a breaking change actually. If a naked function uses `RET` or `RETN` instruction, it must be ensured that the correct value of the FRM register is on the top of the stack since `RET`/`RETN` automatically pop it. Or a manual jump to the return address must be added. I think this is acceptable because the function is marked as naked.

YashasSamaga:
Testing with the current setup is not possible.

Hit LIKE if you approve this PR to be merged into `dev`; otherwise hit DISLIKE.

YashasSamaga:
Would the following lines (line no ~2300) from `sc1.c` need to be changed?

```
if (curfunc->x.stacksize<declared+1)
        curfunc->x.stacksize=declared+1;  /* +1 for PROC opcode */
```

AGraber:
Damn, I’ll take a look when I can regarding that

Issue #333, opened 2018-07-01
We Need a Branching Model


Southclaws:
A few times, new hot fixes to bugs have introduce new bugs because the master branch is cluttered with untested new features. I've proposed git-flow in the past but I really think it's worth consideration now.

See:

- https://nvie.com/posts/a-successful-git-branching-model/
- https://datasift.github.io/gitflow/IntroducingGitFlow.html
- https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow

This should be documented in the `CONTRIBUTING.md` guide as well as in the Pull Request template so contributors are aware of this model.

There are 4 use cases here that I've outlined below on how this branching model should be done:

- External user wants to submit a fix/new feature
- Team member wants to submit a fix/new feature
- External user needs to submit a hotfix to the master branch
- Team member needs to submit a hotfix to the master branch

Here's a rough outline of the workflow for each of those use cases:

- New feature is not urgent
  - if the user is an external contributor:
    - they first fork the repository
    - they then checkout `origin/dev`
    - they then `checkout -b feature-my-new-thing`
    - then work on the code
    - submit a PR that merges `user/feature-my-new-thing` > `compiler/dev`
    - work on the code, review, receive fixes, etc
    - branch is merged to `dev` branch
  - if the user is part of the team:
    - `git checkout dev`
    - `git checkout -b feature-my-new-thing`
    - submit a PR that merges `user/feature-my-new-thing` > `compiler/dev`
    - work on the code, review, receive fixes, etc
    - branch is merged to `dev` branch
- Submission is urgent fix to `master` branch
  - if the user is an external contributor:
    - they first fork the repository
    - they are already on the `master` branch
    - `git checkout -b hotfix-issue-123`
    - work on the code
    - submit a PR that merges `/user/hotfix-issue-123` to `compiler/master`
    - work on the code, review, receive fixes, etc
    - branch is merged to `master` branch
    - a new release is made with the tag `x.y.z-hotfix-123`
  - if the user is part of the team:
    - `git checkout master`
    - `git checkout -b hotfix-issue-123`
    - submit a PR that merges `user/hotfix-issue-123` > `compiler/master`
    - work on the code, review, receive fixes, etc
    - branch is merged to `master` branch

Jeroen52:
I am totally for a new branching model. I have used a modified version of git-flow for a long while like this.

* Master (stable)
    * Hotfix
* Dev (development)
    * Feature
[...]

PR #336, opened 2018-07-03
run clang-format across all compiler source files

Southclaws:

Okay, we don't *need* this, I was mainly just curious what people's opinions were on formatting to be honest because from chatting in Discord, the vibe doesn't seem positive about 2-space indents and extremely tight line spacing! This would most definitely screw up ALL pull request merges so consider it just an idea.

And if people do care about formatting, it would make sense to get all pulls resolved first, organise #333 then maybe come back to this.

I just used WebKit rules, but https://clangformat.com/ provides a nice place to customise a `.clang-format` configuration.

### DIFF
--- a/source/compiler/hashtable/hashtable.h
+++ b/source/compiler/hashtable/hashtable.h
@@ -14,29 +14,27 @@ before you include this file in *one* C/C++ file to create the implementation.
 #define hashtable_h
 
 #ifndef HASHTABLE_U64
-    #define HASHTABLE_U64 unsigned long long
+#define HASHTABLE_U64 unsigned long long
 #endif
 
 typedef struct hashtable_t hashtable_t;
 
-void hashtable_init( hashtable_t* table, int item_size, int initial_capacity, void* memctx );
-void hashtable_term( hashtable_t* table );
+void hashtable_init(hashtable_t* table, int item_size, int initial_capacity, void* memctx);
+void hashtable_term(hashtable_t* table);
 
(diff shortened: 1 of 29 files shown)


Y-Less:
The compiler might not be people's favourite style, but at least it is consistent. This is just going to cause arguments about what it should change to - you already know my view on K&R (which seems especially relevant when you said this was because people thought the code was squashed).

Maybe fine to make sure all places are consistent, but consistent with the style defined in the project, and not your campaign to abolish all existing conventions.

Southclaws:
don't worry, there's no "campaign to abolish all existing conventions", not sure where you got that from. I'm actually an advocate for standard practices and conventions (I often mention the term "convention over configuration" and build things with that in mind).

check out [clang-format](https://clangformat.com/) it's a handy tool, like I said this is purely out of interest comparing the ancient code against modern standards, the clang-format rules I've used are not mine, they are from WebKit - there are also rules used by popular C++ projects such as Chromium, LLVM and a few others (it's a shame C++ hasn't agreed on a single format, but oh well).

and again, not intending to ever merge this for the reasons mentioned in the PR notes, just out of interest and because of a discussion a few of us had a while back!

Y-Less:
Yeah, that was maybe a bit harsh sorry.  Not really a "campaign" I guess, more just that this had a convention that was being changed, and pawn also had a convention that you were trying to change (but your clang-format for pawn has gone for the existing convention as well, so that's good).

PR #345, opened 2018-07-29
update issue template

YashasSamaga:
In my opinion, the current issue template asks for redundant information which bloats the issue with unnecessary details. For example, it has a checklist asking whether the issue is a bug report or a question or a feature request. This is pretty much redundant once the labels are assigned.

The proposed issue template provides a basic concise format and gives more freedom to the reporter to decide on the details of the format..

* [ ] A Bug Fix
* [ ] A New Feature
* [x] Some repository meta (documentation, etc)
* [ ] Other

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,36 +1,37 @@
+<!-- pawn-lang compiler issue template -->
+
 <!--
-This area is for:
-- Bug reports
-- Feature suggestions
-- Questions directly regarding the compiler's implementation, command arguments
-  or usage. This is NOT a place to ask questions about compiler errors or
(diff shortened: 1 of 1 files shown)


Zeex:
I have to agree, big issue templates are scary, when I want to report a bug to some project and their issue template is like 5 pages long I want to just run away. 

But on the other hand it makes think through the problem sometimes and maybe solve it on your own while you are filling this huge form for half an hour....

Southclaws (approved):

Southclaws:
Should this be merged to `master` intead? I'm not sure how GitHub handles branches in relation to issue templates - I would assume it only cares about the version on `master` branch.

Since this is nothing to do with code, there would be no point in waiting for `dev` to be merged to `master` just for this.

Issue #307, opened 2018-05-31
Text bug


Romz24:
I updated the compiler to the latest version (3.10.7), and now the text is not displayed correctly. When I used the old version (3.10.6), there was no problem.

![image](https://user-images.githubusercontent.com/2458113/40813305-c2947a9c-6542-11e8-9292-03a95af26ac0.png)

Southclaws:
You should avoid using Pawno because it mixes tabs and spaces and there's no way to fix it or even know it is happening. What's going on here is the space to the left of `Green Belt` is indented with spaces but then, thanks to Pawno, `Boxing Club` has been switched to tab indentation. In Pawno, these are the same length but in the dialog, tabs are a lot bigger than spaces so that's why you see the difference.

I'd highly advise using either:

- [vscode](http://forum.sa-mp.com/showthread.php?t=647134)
- [Sublime](http://forum.sa-mp.com/showthread.php?t=531379)
- [Atom](http://forum.sa-mp.com/showthread.php?t=549547)

YashasSamaga:
@Romz24 Can you upload the exact string here?

Romz24:
For me this is not a problem, just decided to report just in case. Thank you for answering, you can close the issue.

PR #352, opened 2018-08-05
3.10.8

Zeex:
I'm not sure why this is still not in master, just trying to merge it there.


YashasSamaga:
The 3.10.8 version was a release to fix the bug involving multi-line string concatenation. 

Issue #307 and related issues.

I was wondering if we could revert the offending commit in the `dev` and continue with 3.10.9 instead of merging these changes into `dev`.

Y-Less:
I think it wasn't in master because all the changes were basically duplicated (expect the version number), because it was forked from an earlier commit and had some things back-ported.  So basically everything that is in this was already in master.  But I'll have to look carefully, there may be some things I missed.

Southclaws:
Squashed and merged 👍

PR #353, opened 2018-08-07
Show suggestions for mistyped identifiers

Daniel-Cortez:

Backported this feature from Pawn 3.3; it makes the compiler try to find closest names and suggest them on errors related to undefined (mistyped) identifiers. The current limit for finding close names is 2 mistyped symbols.

Here are some quick examples:
```Pawn
#include <string>
main()
{
    new var = 1;
    printf("%d", val); // error 017: undefined symbol "val"; did you mean "var"?

    printf("%d", celmax); // error 017: undefined symbol "celmax"; did you mean "cellmax"?

    new str[4] = "a";
    strcaf(str, "b"); // error 017: undefined symbol "strcaf"; did you mean "strcat"?

lbl:
    goto lb; // error 019: not a label: "lb"; did you mean "lbl"?
}
```
```Pawn
stock DialogHandler() <dialog:DLG_REGISTER> {}
stock DialogHandler() <dialog:DLG_LOGIN> {}
main()
{
    // error 086: unknown automaton "dialpg"; did you mean "dialog"?
    state dialpg:DLG_REGISTER;

    // error 087: unknown state "DLGLOGIN" for automaton "dialog"; did you mean "DLG_LOGIN"?
    state dialog:DLGLOGIN;
}
```

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -515,7 +515,7 @@ static int find_closestsymbol_table(const char *name,const symbol *root,int symb
       ident=iARRAY;     /* reference arrays match arrays */
     else if (symboltype==iVARIABLE && (sym->ident==iCONSTEXPR || sym->ident==iREFERENCE || sym->ident==iARRAY || sym->ident==iREFARRAY))
       ident=iVARIABLE;  /* when requesting variables, constants are also ok */
-    if ((symboltype==ident || (symboltype==iVARIABLE && ident==iFUNCTN)) && ((ident!=iFUNCTN && ident!=iLABEL) || (sym->usage & uDEFINE)!=0)) {
+    if ((symboltype==ident) && ((ident!=iFUNCTN && ident!=iLABEL) || (sym->usage & uDEFINE)!=0)) {
       dist=levenshtein_distance(name,symname);
       if (dist<closestdist && dist<=maxdist) {
         *closestsym=sym;
@@ -616,7 +616,7 @@ static constvalue *findclosest_automaton_for_state(const char *statename,int fsa
         closestdist=dist;
         if (closestdist<=1)
           break;
-      } /*if */
+      } /* if */
     } /* if */
     ptr=ptr->next;
   } /* while */
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -501,31 +501,42 @@ static int find_closestsymbol_table(const char *name,const symbol *root,int symb
 {
   int dist,maxdist,closestdist=INT_MAX;
   char symname[2*sNAMEMAX+16];
-  symbol *sym=root->next;
+  symbol *sym;
   int ident;
-
   assert(closestsym!=NULL);
   *closestsym=NULL;
   assert(name!=NULL);
   maxdist=get_maxdist(name);
-  while (sym!=NULL) {
-    funcdisplayname(symname,sym->name);
+  for (sym=root->next; sym!=NULL; sym=sym->next) {
     ident=sym->ident;
-    if (symboltype==iARRAY && ident==iREFARRAY)
-      ident=iARRAY;     /* reference arrays match arrays */
-    else if (symboltype==iVARIABLE && (sym->ident==iCONSTEXPR || sym->ident==iREFERENCE || sym->ident==iARRAY || sym->ident==iREFARRAY))
-      ident=iVARIABLE;  /* when requesting variables, constants are also ok */
-    if ((symboltype==ident) && ((ident!=iFUNCTN && ident!=iLABEL) || (sym->usage & uDEFINE)!=0)) {
-      dist=levenshtein_distance(name,symname);
-      if (dist<closestdist && dist<=maxdist) {
-        *closestsym=sym;
-        closestdist=dist;
-        if (closestdist<=1)
-          break;
-      } /* if */
+    if (symboltype==essNONLABEL) {
+      if (ident==iLABEL)
+        continue;
+    } else if (symboltype==essVARCONST) {
+      if (ident!=iCONSTEXPR && ident!=iVARIABLE && ident!=iREFERENCE && ident!=iARRAY && ident!=iREFARRAY)
+        continue;
+    } else if (symboltype==essARRAY) {
+      if (ident!=iARRAY && ident!=iREFARRAY)
+        continue;
+    } else if (symboltype==essCONST) {
+      if (ident!=iCONSTEXPR)
+        continue;
+    } else if (symboltype==essFUNCTN) {
+      if ((ident!=iFUNCTN && ident!=iREFFUNC) || (sym->usage & uDEFINE)==0)
+        continue;
+    } else if (symboltype==essLABEL) {
+      if (ident!=iLABEL || (sym->usage & uDEFINE)==0)
+        continue;
     } /* if */
-    sym=sym->next;
-  } /* while */
+    funcdisplayname(symname,sym->name);
+    dist=levenshtein_distance(name,symname);
+    if (dist>maxdist || dist>=closestdist)
+      continue;
+    *closestsym=sym;
+    closestdist=dist;
+    if (closestdist<=1)
+      break;
+  } /* for */
   return closestdist;
 }
 
@@ -639,7 +650,7 @@ SC_FUNC int error_suggest(int number,const char *name,const char *name2,int type
     if (type!=estSYMBOL) {
       extern char *sc_tokens[];
       name=sc_tokens[subtype-tFIRST];
-      subtype=iVARIABLE;
+      subtype=essVARCONST;
     } /* if */
     closestsym=find_closestsymbol(name,subtype);
     if (closestsym!=NULL)
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -508,6 +508,8 @@ static int find_closestsymbol_table(const char *name,const symbol *root,int symb
   assert(name!=NULL);
   maxdist=get_maxdist(name);
   for (sym=root->next; sym!=NULL; sym=sym->next) {
+    if (sym->fnumber!=-1 && sym->fnumber!=fcurrent)
+      continue;
     ident=sym->ident;
     if (symboltype==essNONLABEL) {
       if (ident==iLABEL)
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -199,7 +199,7 @@ static char *warnmsg[] = {
 };
 
 static char *noticemsg[] = {
-/*001*/  "did you mean \"%s\"?"
+/*001*/  "; did you mean \"%s\"?\n"
 };
 
 #define NUM_WARNINGS    (sizeof warnmsg / sizeof warnmsg[0])
@@ -239,7 +239,7 @@ static short lastfile;
    * "notice" number
    */
   notice=(unsigned long)number >> (sizeof(long)*4);
-  number&=((unsigned long)~0) >> (sizeof(long)*4);
+  number&=(~(unsigned long)0) >> (sizeof(long)*4);
   assert(number>0 && number<300);
 
   /* errflag is reset on each semicolon.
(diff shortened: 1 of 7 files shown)


Daniel-Cortez:
OK, I added tests for this PR. The test code is relatively simple, I just copy-pasted the code from the 1'st comment and added a new case just to make sure the compiler wouldn't output a suggestion if the mistyped name differs from the closest one by more than 2 symbols.

Regarding the other tests, I don't think they need any modifications as this PR already passes all of them just fine.

Southclaws:
This one seems quite large, it would be good to get a full code review from an experienced member. Also one of the tests on AppVeyor has apparently got stuck and GitHub provides no link to it for some reason.

Southclaws:
Restarted https://ci.appveyor.com/project/Southclaws/compiler/build/172

source/compiler/sc5.c
```diff
@@ -197,6 +198,10 @@ static char *warnmsg[] = {
 /*239*/  "literal array/string passed to a non-const parameter\n"
 };
 
+static char *noticemsg[] = {
+/*001*/  "did you mean \"%s\"?"
```
Y-Less: Would `"; did you mean \"%s\"?\n"` be better?  This would be more in-line with the convention in all the other error/warning messages.

source/compiler/sc5.c
```diff
@@ -244,26 +258,39 @@ static short lastfile;
...
     pre=prefix[2];
     warnnum++;
   } /* if */
 
+  if (notice!=0) {
+    size_t len;
+    assert(notice>0 && notice<(1+arraysize(noticemsg)));
+    strcpy(string,msg);
+    len=strlen(string);
+    sprintf(&string[len-1],"; %s\n",noticemsg[notice-1]);
```
Y-Less: Were the change to `noticemsg` made, this would become just `strcat`.

source/compiler/sc5.c
```diff
@@ -220,13 +225,22 @@ static int errwarn;
...
   char *msg,*pre;
   va_list argptr;
+  char string[128];
+  int notice;
+
+  /* split the error field between the real error/warning number and an optional
+   * "notice" number
+   */
+  notice=(unsigned long)number >> (sizeof(long)*4);
+  number&=((unsigned long)~0) >> (sizeof(long)*4);
```
Y-Less: Better as `~0UL >> sizeof (long) * 4`?

source/compiler/sc5.c
```diff
@@ -423,3 +450,173 @@ int pc_geterrorwarnings()
...
+  } else if (type==estSTATE) {
+    constvalue *closestautomaton=find_closeststate(name,subtype);
+    if (closestautomaton!=NULL)
+      closestname=closestautomaton->name;
+  } else {
+    assert(0);
+  } /* if */
+
+  if (closestname!=NULL) {
+    if (name2!=NULL)
```
Y-Less: Swap the conditions around a `not` condition with an `else` is just confusing.

Y-Less:
```c
  if (closestname==NULL) {
    error(number,name,name2);
  } else if (name2==NULL) {
    error(makelong(number,1),name,closestname);
  } else {
    error(makelong(number,1),name,name2,closestname);
  }
```

source/compiler/sc5.c
```diff
@@ -423,3 +450,173 @@ int pc_geterrorwarnings()
...
+  int dist,closestdist=INT_MAX;
+
+  assert(name!=NULL);
+  while (ptr!=NULL) {
+    if (strlen(ptr->name)>0) {
+      dist=levenshtein_distance(name,ptr->name);
+      if (dist<closestdist && dist<=MAX_EDIT_DIST) {
+        closestmatch=ptr;
+        closestdist=dist;
+        if (dist<=1)
```
Y-Less: There's a lot of repetitive code here, which is never great, but I do see why - there are very subte differences in things like the loop condition, for example `find_closeststate` checking the automata is correct so as not to suggest matching states from other ones.

One difference I'm not clear on is this `break`.  It looks like a nice optimisation - if there is a very similar symbol, suggest that and don't bother to continue searching.  However, it is only in this one function, not all of them (which is, of course, the danger with repeated code).

source/compiler/sc5.c
```diff
@@ -423,3 +450,173 @@ int pc_geterrorwarnings()
...
+static constvalue *find_closeststate(const char *name,int fsa)
+{
+  constvalue *ptr=sc_state_tab.first;
+  constvalue *closestmatch=NULL;
+  int dist,closestdist=INT_MAX;
+
+  assert(name!=NULL);
+  assert(closestmatch!=NULL);
+  while (ptr!=NULL) {
+    if (ptr->index==fsa && strlen(ptr->name)>0) {
```
Y-Less: It would be nice if `find_closestautomaton` and this could be merged - is there a constant value of `->index` in `sc_automaton_tab`?  If so, `fsa` could be set to that, and an extra function deleted.

source/compiler/tests/gh_353.meta
```diff
@@ -0,0 +1,13 @@
...
+  'errors': """
+gh_353.pwn(6) : error 017: undefined symbol "var"
+gh_353.pwn(9) : error 017: undefined symbol "val"; did you mean "var"?
+gh_353.pwn(11) : error 017: undefined symbol "celmax"; did you mean "cellmax"?
+gh_353.pwn(14) : error 017: undefined symbol "strcaf"; did you mean "strcat"?
+gh_353.pwn(20) : error 086: unknown automaton "dialpg"; did you mean "dialog"?
+gh_353.pwn(20) : error 036: empty statement
+gh_353.pwn(23) : error 087: unknown state "DLGLOGIN" for automaton "dialog"; did you mean "DLG_LOGIN"?
+gh_353.pwn(23) : error 036: empty statement
+  """
```
Y-Less: No tests for errors 19, 20, 28, or 80.

source/compiler/tests/gh_353.pwn
```diff
@@ -0,0 +1,26 @@
...
+lbl:
+	goto lb; // error 019: not a label: "lb"; did you mean "lbl"?
+
+	// error 086: unknown automaton "dialpg"; did you mean "dialog"?
+	state dialpg:DLG_REGISTER;
+
+	// error 087: unknown state "DLGLOGIN" for automaton "dialog"; did you mean "DLG_LOGIN"?
+	state dialog:DLGLOGIN;
+}
+stock DialogHandler() <dialog:DLG_REGISTER> {}
```
Y-Less: Would also be nice to have a test case something like:

```pawn
Func1() <state_1:STATE_1>{}
Func1() <state_1:STATE_2>{}

Func2() <state_2:BEING_1>{}
Func2() <state_2:BEING_2>{}

state state_2:STATE_2;
```

Which should NOT suggest an alternative, because `STATE_2` is in a different automata.  The code looks like it does this, but confirmation would be nice.

Or even better, in that case suggest that maybe you meant `state_1:STATE_2`, which the code currently does not do.

Daniel-Cortez:
> it would be good to get a full code review from an experienced member

By the way, I believe it's worth to note that the code in this PR isn't exactly backported - in fact the only things I copied from Pawn 3.3 were the `noticemsg` array, function `find_closestsymbol_table()` (with a few minor modifications to remove false positives), the Levenshtein distance implementation and maybe a few other simple things (like the `makelong` macro in sc.h and the idea of changing `error(int number,...)` to `error(long number,...)`). All of the other code was either based on the already existing code (f.e. `find_closestautomaton()` and `find_closeststate()` are derived from the code from `find_automaton()` and `find_state()` respectively) or made from scratch.

The original implementation from 3.3 tries to find the closest match even on the first pass which IMO is not the brightest idea because
  1. `error()` doesn't show non-fatal error messages on the first pass anyway (so the same message wouldn't be reported twice);
  2. usually there's a lot of undefined identifiers on the first pass that get resolved by the second pass, so trying to find the closest match for every one of them would be just a waste of time.

And for states and automatons the original code doesn't seem to be well thought out either:
  1. the Levenshtein distance is calculated directly in `find_automaton()` and `find_state()` on every compared name which again slows down compilation for nothing;
  2. instead of having a separate function for suggesting the closest state/automaton name and keeping everything in one place (in sc5.c) the code that tries to find the closest automaton/state names is located in `sc_getstateid()` (sc3.c).

So that's why I partially reimplemented suggestions and kept only the most basic parts of the original code from Pawn 3.3 I mentioned earlier.

source/compiler/sc5.c
Daniel-Cortez: Yeah, looks like I forgot to add that check in the other functions. Thanks for noting this.

source/compiler/sc5.c
Daniel-Cortez: I did it this way so there would be more similarity between this line and the above one.

Y-Less: Except casting doesn't sign extend.  So this is:

`(unsigned long)0xFFFFFFFF`

if `sizeof (long) == sizeof (int)` that's fine, if it isn't, you will get `0x00000000FFFFFFFF`.  After shifting that will give 0, not `0xFFFF`.

Daniel-Cortez: Oh, so that's what you meant. Yeah, I already noticed that mistake the first time you left a review comment at that line and fixed it locally by just moving the `~` before `(unsigned long)`. Thanks for pointing it out.

Daniel-Cortez:
Ok, so I made the following changes to the code:

1. The messages for error 020 were altered for the cases when the user types a keyword instead of an identifier. For example, now when the user mistypes "asset" as "assert", the compiler prints `invalid symbol name "assert"` (before this PR it printed `invalid symbol name ""`).

2. If there is an undefined state name with no close names, the compiler attempts to suggest an automaton with a close name.
  Example:
```Pawn
Func1()<automaton_1:STATE_1>{}
Func2()<automaton_2:BEING_1>{}

state automaton_2:STATE_1; // unknown state "STATE_1" for automaton "automaton_2"; did you mean "automaton_1:STATE_1"?
```
3. Rewrote the tests, added test cases for errors 019, 020 and 080 (there's still no test for error 028; see below).

4. Other minor adjustments.

Things I'm still not sure about:

1. I tried to add a test case for error 028
```Pawn
forward test_e028();
public test_e028()
{
	// error 028: invalid subscript (not an array or too many subscripts): "value"; did you mean "values"?
	const value = 0;
	static const values[2] = { 1, 2 };
	printf("%d\n", value[0]);
	#pragma unused values
}
```
but it made the whole test fail, even though the error output was exactly the same as expected, so I removed that case for now. I suspect the test was failing because of a fatal error (`too many error messages on one line`) triggered by that test case. 
```
gh_353.pwn(81) : error 028: invalid subscript (not an array or too many subscripts): "value"; did you mean "values"?
gh_353.pwn(81) : warning 215: expression has no effect
gh_353.pwn(81) : error 001: expected token: ";", but found "]"
gh_353.pwn(81) : error 029: invalid expression, assumed zero
gh_353.pwn(81) : fatal error 107: too many error messages on one line
```
Is there any way to not trigger a fatal error on 080?

2. Previously if the mistyped symbol (variable/constant/function/label name) was too short (3 or less characters), only a difference of 1 character instead of 2 was allowed, so f.e. the compiler wouldn't suggest "value" for "val". For now I also applied the same rule for automatons and states, but maybe it would be better to remove that rule completely (or rather limit it to the names with length of 2 or less characters instead of 3, so f.e. the compiler would suggest "value" for "val" but wouldn't suggest "var0" for "va")?

@Y-Less

Y-Less:
Aren't they all just symbols to the compiler? Though that would be a good thing.

Daniel-Cortez:
> Aren't they all just symbols to the compiler?

If there are parentheses after the symbol, the compiler recognizes a function call and thinks the symbol is an unimplemented function.

> Though that would be a good thing.

OK, done. BTW, GitHub says you're still requesting changes - is there anything else that should be done for this PR? (If it's because there's no test case for error 028, I already explained why I couldn't add it.)

Zeex:
Merged (https://github.com/pawn-lang/compiler/commit/bc04b2ef048a02a7f4030d807d97a7fc10cb760a).

Issue #354, opened 2018-08-10
#pragma unused doesn't work


Zeex:
### Issue description:
Despite `#pragma unused v` this code still produced a "symbol is never used warning":

```pawn
f1(const &v) { }
f2(const ...) { }
f3(const v) {
	#pragma unused v
}
f4(...) { }
f5(v) {
	#pragma unused v
}
f6(&v) { }

main() {
	new a;
	f1(a);
	f2(a);
	f3(a);
	f4(a);
	f5(a);
	f6(a);
}
```

(It's from the `meaningless_class_specifiers_gh_172` test)

### Workspace Information:

* Compiler version: dev branch

Zeex:
Maybe @YashasSamaga knows?

YashasSamaga:
The warnings are correct. There is no `#pragma unused` for `f1` and `f6`. I think I forgot to add mark the symbols as unused in the test script.

Zeex:
Yeah I'm totally blind, didn't see that those pragmas were in other functions

PR #355, opened 2018-08-10
Add a custom test runner script

Zeex:
At the moment it supports only tests that output warnings or errors (test type = output_check). More types can be added in the future.

Advantages over the old CTest system:

* When adding new tests there should not be any conflicts now because each test lives in its own file (or 2 files rather)
* Makes it easier to add other types of tests, e.g. check produced assembly listings or run-time test - with pure CMake it's hard to do
* Don't need to use regular expressions, it's simpler to just copy output from your terminal and paste into a corresponding meta file (support for regexps could be added later if needed)
* Don't need to re-configure the project with CMake after editing tests

Disadvantages:
* Need Python to run the tests (version 3)

Each test has a meta file which is simply a dictionary in Python syntax with some information about the test (its type, expected output and other properties).


Zeex:
BTW I disabled 2 tests because they were failing:

* md_array_size_chk_gh_314
* meaningless_class_specifiers_gh_172 - see #354

PR #357, opened 2018-08-13
[WIP] Check if function is naked before adding +1 for PROC opcide

AGraber:

- Finishes stuff for naked functions that went unnoticed on #335 

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:
This PR isn't finished yet, but I wanted to get some feedback on what would be the proper way to fix this particular situation:

`fetchfunc`:
https://github.com/pawn-lang/compiler/blob/694c47f53b75f9b57ef7aa551ece94664d9c37ea/source/compiler/sc1.c#L3167-L3170

This particular check would always fail (and set the stacksize to 1) since the `flagNAKED` isn't set when `fetchfunc` is called from `newfunc`.

Any ideas?


Southclaws:
LGTM!

PR #363, opened 2018-09-06
amxaux.c: Don't forget to close the program file before returning

Daniel-Cortez:

Fixes bugs introduced in 44312b96532cc1056a9f6e46bec82525d5a53229 (the interpreter won't close the AMX file if `aux_ProgramSize()` or `aux_LoadProgram()` couldn't read the data from it). Not that these bugs matter much (the file is closed automatically by the OS when the interpreter process is terminated anyway), but I believe it still would be right to fix them.

-

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/amx/amxaux.c
+++ b/source/amx/amxaux.c
@@ -30,13 +30,15 @@
 size_t AMXAPI aux_ProgramSize(char *filename)
 {
   FILE *fp;
+  size_t size;
   AMX_HEADER hdr;
 
   if ((fp=fopen(filename,"rb")) == NULL)
     return 0;
-  if (fread(&hdr, sizeof hdr, 1, fp) < 1)
-    return 0;
+  size = fread(&hdr, sizeof hdr, 1, fp);
   fclose(fp);
+  if (size < 1)
+    return 0;
 
   amx_Align16(&hdr.magic);
   amx_Align32((uint32_t *)&hdr.stp);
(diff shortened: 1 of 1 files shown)


source/amx/amxaux.c
```diff
@@ -46,21 +48,24 @@ size_t AMXAPI aux_ProgramSize(char *filename)
...
     return AMX_ERR_NOTFOUND;
-  if (fread(&hdr, sizeof hdr, 1, fp) < 1)
-    return AMX_ERR_INIT;
-  amx_Align16(&hdr.magic);
-  amx_Align32((uint32_t *)&hdr.size);
-  amx_Align32((uint32_t *)&hdr.stp);
-  if (hdr.magic != AMX_MAGIC) {
+  size = fread(&hdr, sizeof hdr, 1, fp);
+  if (size < 1) {
+err_format:
```
Zeex: Can we have the goto here instead of below so that it jumps forward? or remove it at all

Y-Less:
I thought we weren't doing VM changes?

Y-Less:
Though I think we should.

Zeex:
@Y-Less amxaux is now used by [pawnruns](https://github.com/pawn-lang/compiler/blob/dev/source/compiler/pawnruns.c) (the program that runs runtime tests), so I think we do need to take care of bugs related to it. 

I know that I kind of said earlier that VM changes are out of scope of this project - that is because pretty much nobody uses this version of the VM, so if something gets broken hardly anyone would notice. With the compiler it's easy - just compile stuff and see if it works, and there are a lot of people who use it and can report bugs. But with the VM how do you verify that some patch or feature works? 

Well, I'm not the one to decide what to include anyway since I kind of retired, but if VM patches are useful to somebody I don't have a problem with that, feel free to accept them 😃

Y-Less:
I use this version.

Zeex:
OK.

Daniel-Cortez:
Most (if not all) of the bugfixes and security patches to the VM would be rejected anyway, considering how much this project is focused on compatibility with SA-MP (I would gladly donate the patches from my fork of Pawn 3.2 otherwise), so what's the point?

PR #372, opened 2018-10-03
check all definitions of a function before suggesting 'const'

### LINKED ISSUE #371: Impossible to satisfy const-correctness
### Issue description:

When you have two states for a function, one of which modifies an array, the other of which doesn't, the compiler will complain about the const-correctness of the argument regardless of what you do.

```pawn
MyFunc(str[], &val) <my_state : a>
{
	val = str[0];
}

MyFunc(str[], &val) <my_state : b>
{
	str[0] = val;
}

main()
{
	state my_state : a;
	new b, arr[10];
	MyFunc(arr, b);
}
```

`(1) : warning 214: possibly a "const" array argument was intended: "str"`

```pawn
MyFunc(const str[], &val) <my_state : a>
{
	val = str[0];
}

MyFunc(str[], &val) <my_state : b>
{
	str[0] = val;
}

main()
{
	state my_state : a;
	new b, arr[10];
	MyFunc(arr, b);
}
```

`(6) : error 025: function heading differs from prototype`

```pawn
MyFunc(const str[], &val) <my_state : a>
{
	val = str[0];
}

MyFunc(const str[], &val) <my_state : b>
{
	str[0] = val;
}

main()
{
	state my_state : a;
	new b, arr[10];
	MyFunc(arr, b);
}
```

`(8) : error 022: must be lvalue (non-constant)`
`(8) : warning 215: expression has no effect`

```
```

I did discover that you can shut the warning up with this:

```pawn
MyFunc(str[], &val) <my_state : a>
{
	#pragma unused str
	val = str[0];
}

MyFunc(str[], &val) <my_state : b>
{
	str[0] = val;
}

main()
{
	state my_state : a;
	new b, arr[10];
	MyFunc(arr, b);
}
```

I don't know if the fact that that works is a bug in itself...
### Workspace Information:

* Compiler version:
* Operating System:

YashasSamaga:

`const` qualifier suggestions were being given based solely on the function definition that was currently being looked at. This caused wrong suggestions when one state of a function modifies the argument while another state did not.

The commit in this PR adds code to share usage details across multiple definitions of the functions. Hence, when the `const` qualifier is suggested, it takes into account of all the function states.

The commit is just a  proof-of-concept for the time being. The commit only shares `uWRITTEN` status only but it could be generalized to share other usage flags as well which could solve https://github.com/pawn-lang/compiler/issues/371#issuecomment-426589370

Fixes #371 

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -108,7 +108,7 @@ static void funcstub(int fnative);
 static int newfunc(char *firstname,int firsttag,int fpublic,int fstatic,int stock);
 static int declargs(symbol *sym,int chkshadow);
 static void doarg(char *name,int ident,int offset,int tags[],int numtags,
-                  int fpublic,int fconst,int chkshadow,arginfo *arg);
+                  int fpublic,int fconst,int written,int chkshadow,arginfo *arg);
 static void make_report(symbol *root,FILE *log,char *sourcefile);
 static void reduce_referrers(symbol *root);
 static long max_stacksize(symbol *root,int *recursion);
@@ -3680,8 +3680,8 @@ static void funcstub(int fnative)
  */
 static int newfunc(char *firstname,int firsttag,int fpublic,int fstatic,int stock)
 {
-  symbol *sym;
-  int argcnt,tok,tag,funcline;
+  symbol *sym,*lvar,*depend;
+  int argcnt,tok,tag,funcline,i;
   int opertok,opererror;
   char symbolname[sNAMEMAX+1];
   char *str;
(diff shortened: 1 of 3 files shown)


YashasSamaga:
Ready.

Daniel-Cortez:
This seems like a useful bugfix, as it saves from the need to manually apply `#pragma unused` as a workaround in all function states that don't modify the argument.
@YashasSamaga Why did you close this PR?

YashasSamaga:
I lost touch with the code and I planned on writing a generic system for any flag instead of just `uWRITTEN` but never got time to do it.

Daniel-Cortez:
Hmm... I see what you mean; a few functions in `sc1.c` indeed do have an unusually high number of arguments. But implementing a flag system might complicate the code even more. And even if it won't, does it really have to be done in this PR, and not in a separate one later (after this and the `{...}` PRs are merged)?

PR #374, opened 2018-10-03
Fix the compiler suggesting unimplemented variables

Daniel-Cortez:

This PR fixes the problem where after the 1'st pass the compiler suggests global variables before the point of their definition:
```Pawn
GetValue()
{
    return value; // error 017: undefined symbol "value"; did you mean "value"?
}

new value = 0;

main()
{
    GetValue();
}
```

-

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

-

### DIFF
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -510,6 +510,8 @@ static int find_closest_symbol_table(const char *name,const symbol *root,int sym
   for (sym=root->next; sym!=NULL; sym=sym->next) {
     if (sym->fnumber!=-1 && sym->fnumber!=fcurrent)
       continue;
+    if ((sym->usage & uDEFINE)==0)
+      continue;
     ident=sym->ident;
     if (symboltype==essNONLABEL) {
       if (ident==iLABEL)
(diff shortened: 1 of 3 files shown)


YashasSamaga:
There is an `assert(0)` in `error_suggest` but doing `tagof(X)` where `X` is a string, number, etc. leads to that assertion because of `error_suggest(20,st,NULL,estNONSYMBOL,tok);`. Since `tok` is passed as `subtype`, it can be `tRATIONAL`, `tNUMBER`, `tSTRING`, etc. depending on what the user entered.

The previous version of the compiler was giving out errors instead of an assertion failure.

Y-Less:
I have wished you could take the tag of literals a few times:

`tagof (X:5)`

Not obviously useful as typed, but is in macro outputs.  But I have macros to remove anything after the `:` for now.

Daniel-Cortez:
Ok, I fixed this and a couple of other bugs locally. Would it be alright if I turn this PR into a more general one and upload those bugfixes here? (Just asking because this PR has already been approved in the current state.)

Y-Less:
No, make those a separate pull request.  The point of approvals is to show that the code's been checked, adding more after that defeats the purpose.

Daniel-Cortez:
Well, I wasn't asking this only because this PR has already been approved, but also because uploading those fixes in a separate PR would most probably lead to a mess with merge conflicts, but ok.

YashasSamaga:
I think every new feature should have a corresponding issue open. I couldn't find an issue for this system which is why I sneaked it in here.

PR #375, opened 2018-10-04
Write `.lst` after the second pass.

### LINKED ISSUE #116: -l bug
Default ALS method:
```Pawn
native printf(const format[], {Float,_}:...);
forward OnGameModeInit();

main() {}

// ---------------------------------------------------------------------
public OnGameModeInit()
{
	printf("OnGameModeInit 1");
	#if defined Prefix_OnGameModeInit
		return Prefix_OnGameModeInit();
	#else
		return 1;
	#endif
}
#if defined _ALS_OnGameModeInit
	#undef OnGameModeInit
#else
	#define _ALS_OnGameModeInit
#endif

#define OnGameModeInit Prefix_OnGameModeInit
#if defined Prefix_OnGameModeInit
	forward Prefix_OnGameModeInit();
#endif

// ---------------------------------------------------------------------
public OnGameModeInit()
{
	printf("OnGameModeInit 2");
	#if defined Prefix2_OnGameModeInit
		return Prefix2_OnGameModeInit();
	#else
		return 1;
	#endif
}
#if defined _ALS_OnGameModeInit
	#undef OnGameModeInit
#else
	#define _ALS_OnGameModeInit
#endif

#define OnGameModeInit Prefix2_OnGameModeInit
#if defined Prefix2_OnGameModeInit
	forward Prefix2_OnGameModeInit();
#endif

// ---------------------------------------------------------------------
public OnGameModeInit()
{
	printf("OnGameModeInit");
	return 1;
}
```
Results:
```
OnGameModeInit 1
OnGameModeInit 2
OnGameModeInit
```
It's ok, but if I compile it with -l param, I got this:
```Pawn
#pragma ctrlchar 0x5c
#pragma pack false
#pragma semicolon true
#pragma tabsize 8
#file "sources\default.pwn"
#line 1

native printf(const format[], {Float,_}:...);

main() {}

forward OnGameModeInit();

public OnGameModeInit()
{
	printf("OnGameModeInit 1");
#line 14
		return 1;
#line 16
}
#line 22

#line 27






public Prefix_OnGameModeInit()
{
	printf("OnGameModeInit 2");
#line 39
		return 1;
#line 41
}
#line 47

#line 52


public Prefix2_OnGameModeInit()
{
	printf("OnGameModeInit");
	return 1;
}
```
And this is wrong.

Y-Less:

This moves the `-l` code output to after the second pass, so that the generated code is the most accurate it can be.
-

Fixes #116

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -280,6 +280,7 @@ enum {
 enum {
   statIDLE,     /* not compiling yet */
   statFIRST,    /* first pass */
+  statSECOND,   /* second pass */
   statWRITE,    /* writing output */
   statSKIP,     /* skipping output */
 };
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -575,20 +575,6 @@ int pc_compile(int argc, char *argv[])
       fline++;                  /* keep line number up to date */
   skipinput=fline;
   sc_status=statFIRST;
-  /* write starting options (from the command line or the configuration file) */
-  if (sc_listing) {
-    char string[150];
-    sprintf(string,"#pragma ctrlchar 0x%02x\n"
-                   "#pragma pack %s\n"
-                   "#pragma semicolon %s\n"
-                   "#pragma tabsize %d\n",
-            sc_ctrlchar,
-            sc_packstr ? "true" : "false",
-            sc_needsemicolon ? "true" : "false",
-            sc_tabsize);
-    pc_writeasm(outf,string);
-  } /* if */
-  setfiledirect(inpfname);
   /* do the first pass through the file (or possibly two or more "first passes") */
   sc_parsenum=0;
   inpfmark=pc_getpossrc(inpf_org);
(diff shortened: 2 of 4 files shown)


Southclaws:
I might revive my parser project because of this!

AGraber:
I think there should be a way to still print the first pass, through another flag of course.

Southclaws:
Yes that would be useful for investigating exactly what the difference is with complex macros.

Y-Less:
The only difference this makes to macros is code like this:

```pawn
#if defined Hooked_OnPlayerConnect
    Hooked_OnPlayerConnect(playerid);
#endif
```

All other macros, unless you are doing very odd things with future definitions, will be the same.  It doesn't replace at a later stage in macro replacement, those have always been atomic.

Southclaws:
Code looks fine to my (untrained) eyes, since I intend to work on a parser soon I wanted to get this in for the next version. I'll hopefully get around to testing this at the weekend.

PR #376, opened 2018-10-04
Branch pragma deprecated

### LINKED ISSUE #242: redundant pragma deprecated warning at the site of declaration
```
#pragma deprecated don't use please
f() {

}

main() {
  
}
```
no deprecated function warning for the above code which is right
```
#pragma deprecated don't use please
f() {

}

main() {
  f();
}
```
two warnings are thrown instead of just one

I think this is not right. : the warning thrown at the site of declaration of the deprecated function is redundant.

Y-Less:

Hides the first deprecation warning, so that only the use-site is warned about, not the declaration site.

Fixes #242 

* [ ] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -3774,10 +3774,6 @@ static int newfunc(char *firstname,int firsttag,int fpublic,int fstatic,int stoc
     cidx=code_idx;
     glbdecl=glb_declared;
   } /* if */
-  if ((sym->flags & flgDEPRECATED)!=0) {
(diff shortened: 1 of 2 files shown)


YashasSamaga:
I think the code you removed is actually not redundant: https://github.com/pawn-lang/compiler/issues/242#issuecomment-355860243

Y-Less:
I think this is already in from someone else.

PR #393, opened 2018-12-24
use enum for defining tokens

YashasSamaga:

The old system used a series of `#define`s to assign a value to each token. The compiler does not use the values directly but requires the tokens values to be unique.

The problem arises when a new token has to be added. The tokens are sorted alphabetically. To maintain the consistency of the order, the developer needs to add his token somewhere in the middle (in most cases) and offset the rest of the entries by one. Not just that, the developer would also have to update `tLAST`. This is highly error-prone.

With this PR, the developer would just have to insert his token somewhere and the remaining values would automatically adjust themselves.

* [ ] A Bug Fix
* [ ] A New Feature
* [x] Other

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1082,7 +1082,7 @@ static int command(void)
     } /* if */
     check_empty(lptr);
     break;
-  case tINCLUDE:                /* #include directive */
+  case tpINCLUDE:                /* #include directive */
   case tpTRYINCLUDE:
     ret=CMD_INCLUDE;
     if (!SKIPPING)
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -2133,7 +2133,7 @@ SC_FUNC void lexinit(void)
   _lexnewline=FALSE;
 }
 
-char *sc_tokens[] = {
+char *sc_tokens[tlexLAST-tFIRST+1] = {
          "*=", "/=", "%=", "+=", "-=", "<<=", ">>>=", ">>=", "&=", "^=", "|=",
          "||", "&&", "==", "!=", "<=", ">=", "<<", ">>>", ">>", "++", "--",
          "...", "..",
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -2133,7 +2133,7 @@ SC_FUNC void lexinit(void)
   _lexnewline=FALSE;
 }
 
-char *sc_tokens[tlexLAST-tFIRST+1] = {
+char *sc_tokens[] = {
          "*=", "/=", "%=", "+=", "-=", "<<=", ">>>=", ">>=", "&=", "^=", "|=",
          "||", "&&", "==", "!=", "<=", ">=", "<<", ">>>", ">>", "++", "--",
          "...", "..",
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -2136,7 +2136,7 @@ SC_FUNC void lexinit(void)
 char *sc_tokens[] = {
          "*=", "/=", "%=", "+=", "-=", "<<=", ">>>=", ">>=", "&=", "^=", "|=",
          "||", "&&", "==", "!=", "<=", ">=", "<<", ">>>", ">>", "++", "--",
-         "...", "..", "::",
+         "...", "..",
          "assert", "*begin", "break", "case", "char", "const", "continue", "default",
          "defined", "do", "else", "emit", "__emit", "*end", "enum", "exit", "for",
          "forward", "goto", "if", "native", "new", "operator", "public", "return",
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -345,9 +345,8 @@ enum {
   tDEC,   /* -- */
   tELLIPS,  /* ... */
   tDBLDOT,  /* .. */
-  tDBLCOLON, /* :: */
 
-  tMIDDLE = tDBLCOLON, /* value of last multi-character operator */
+  tMIDDLE = tDBLDOT, /* value of last multi-character operator */
 
   /* reserved words (statements) */
   tASSERT,
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -421,10 +421,7 @@ enum {
   teLOCAL, /* local variable (name or offset) */
   teFUNCTN, /* Pawn function */
   teNATIVE, /* native function */
-  teNONNEG, /* nonnegative integer */
-
-  tlexLAST = teNONNEG,
-
+  teNONNEG , /* nonnegative integer */
   /* for assigment to "lastst" only (see SC1.C) */
   tEXPR,
   tENDLESS, /* endless loop */
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -421,7 +421,10 @@ enum {
   teLOCAL, /* local variable (name or offset) */
   teFUNCTN, /* Pawn function */
   teNATIVE, /* native function */
-  teNONNEG , /* nonnegative integer */
+  teNONNEG, /* nonnegative integer */
+
+  tlexLAST = teNONNEG,
+
   /* for assigment to "lastst" only (see SC1.C) */
   tEXPR,
   tENDLESS, /* endless loop */
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -396,7 +396,7 @@ enum {
   tpERROR,
   tpFILE,
   tpIF,/* #if */
-  tINCLUDE,
+  tpINCLUDE,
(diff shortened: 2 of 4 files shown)


YashasSamaga:
Should `tDBLCOLON` be removed as it is unused?

Daniel-Cortez:
Didn't you already do this in #300? What happened to that PR?

> Should tDBLCOLON be removed as it is unused?

Probably, it's not like we are going to transform the language into something like SourcePawn anyway.

Southclaws (approved):

Daniel-Cortez:
@YashasSamaga While you're at it, could you please also rename `tINCLUDE` to `toINCLUDE` and `flgDEPRECATED` to `flagDEPRECATED`?
I already described these naming inconsistencies here: https://github.com/pawn-lang/compiler/issues/296#issuecomment-406837564

YashasSamaga:
I am surprised how removing `tDBLCOLON` broke the entire compiler. I suspect hardcoding of constants somewhere. I need to search.

YashasSamaga:
To test I deleted the token `tDBLCOLON` in the original code and it broke the compiler. This appears to be far more serious. I am not sure why it breaks and I don't know how to even start searching.

```
#define tFIRST      256 /* value of first multi-character operator */
#define tMIDDLE     279 /* value of last multi-character operator */
#define tLAST       330 /* value of last multi-character match-able token */
/* multi-character operators */
#define taMULT      256 /* *= */
#define taDIV       257 /* /= */
#define taMOD       258 /* %= */
#define taADD       259 /* += */
#define taSUB       260 /* -= */
#define taSHL       261 /* <<= */
#define taSHRU      262 /* >>>= */
#define taSHR       263 /* >>= */
#define taAND       264 /* &= */
#define taXOR       265 /* ^= */
#define taOR        266 /* |= */
#define tlOR        267 /* || */
#define tlAND       268 /* && */
#define tlEQ        269 /* == */
#define tlNE        270 /* != */
#define tlLE        271 /* <= */
#define tlGE        272 /* >= */
#define tSHL        273 /* << */
#define tSHRU       274 /* >>> */
#define tSHR        275 /* >> */
#define tINC        276 /* ++ */
#define tDEC        277 /* -- */
#define tELLIPS     278 /* ... */
#define tDBLDOT     279 /* .. */
/* reserved words (statements) */
#define tASSERT     280
#define tBEGIN      281
#define tBREAK      282
#define tCASE       283
#define tCHAR       284
#define tCONST      285
#define tCONTINUE   286
#define tDEFAULT    287
#define tDEFINED    288
#define tDO         289
#define tELSE       290
#define tEMIT       291
#define t__EMIT     292
#define tEND        293
#define tENUM       294
#define tEXIT       295
#define tFOR        296
#define tFORWARD    297
#define tGOTO       298
#define tIF         299
#define tNATIVE     300
#define tNEW        301
#define tOPERATOR   302
#define tPUBLIC     303
#define tRETURN     304
#define tSIZEOF     305
#define tSLEEP      306
#define tSTATE      307
#define tSTATIC     308
#define tSTOCK      309
#define tSWITCH     310
#define tTAGOF      311
#define tTHEN       312
#define tWHILE      313
/* compiler directives */
#define tpASSERT    314 /* #assert */
#define tpDEFINE    315
#define tpELSE      316 /* #else */
#define tpELSEIF    317 /* #elseif */
#define tpEMIT      318
#define tpENDIF     319
#define tpENDINPUT  320
#define tpENDSCRPT  321
#define tpERROR     322
#define tpFILE      323
#define tpIF        324 /* #if */
#define tINCLUDE    325
#define tpLINE      326
#define tpPRAGMA    327
#define tpTRYINCLUDE 328
#define tpUNDEF     329
#define tpWARNING   330
/* semicolon is a special case, because it can be optional */
#define tTERM       331 /* semicolon or newline */
#define tENDEXPR    332 /* forced end of expression */
/* other recognized tokens */
#define tNUMBER     333 /* integer number */
#define tRATIONAL   334 /* rational number */
#define tSYMBOL     335
#define tLABEL      336
#define tSTRING     337
/* argument types for emit/__emit */
#define teANY       338 /* any value */
#define teNUMERIC   339 /* integer/rational number */
#define teDATA      340 /* data (variable name or address) */
#define teLOCAL     341 /* local variable (name or offset) */
#define teFUNCTN    342 /* Pawn function */
#define teNATIVE    343 /* native function */
#define teNONNEG    344 /* nonnegative integer */
/* for assigment to "lstst" only (see SC1.C) */
#define tEXPR       345
#define tENDLESS    346 /* endless loop */
```

Y-Less:
Could start with a search for `\d{3}`.

YashasSamaga:
I was thinking of adding `undecided1` and `undecided2` tokens which mark the first and the last token processed by the lexer. This can be used here: `sc_token[undecided2 - undecided1+ 1]`. This would catch people who forget to update the token strings list after adding/removing a token.

A better solution could be possible. There should be a way to link `sc_tokens` and the token enumerator. The issue is that they are completely delinked and the order in which the tokens are specified in `sc_tokens` must exactly match the constants assigned but there is currently no way to enforce this.

Daniel-Cortez:
> I was thinking of adding undecided1 and undecided2 tokens which mark the first and the last token processed by the lexer.

We already have `tFIRST`, so I think we need only one new enum element for the last string in `sc_tokens`.
```C
  teFUNCTN, /* Pawn function */
  teNATIVE, /* native function */
  teNONNEG , /* nonnegative integer */

  tUNDECIDED = teNONNEG,

  /* for assigment to "lastst" only (see SC1.C) */
  tEXPR,
  tENDLESS, /* endless loop */
};
```
Also, there's macro `assert_static` defined in `amx.h`, we can use it to trigger a compile-time error when the number of tokens in the token enum doesn't match with the number of strings in `sc_tokens`. And due to how `assert_static` is implemented, this check will only work inside a function, so we could place it, say, at the very beginning of function `lex()`:
```C
SC_FUNC int lex(cell *lexvalue,char **lexsym)
{
  int i,toolong,newline,stringflags;
  char **tokptr;
  const unsigned char *starttoken;

  assert_static(sizeof sc_tokens/sizeof sc_tokens[0]==tUNDECIDED-tFIRST+1);

  // ...
```

YashasSamaga:
I have used `tlexLAST` which isn't a great name (if not bad) but it enforces the size requirement.

YashasSamaga:
I think this PR has quite a lot of improvements and should be merged so that future PRs utilize these changes.

It should be easy to review the code as it contains very basic changes.

Southclaws (approved):

source/compiler/sc2.c
```diff
@@ -1082,7 +1082,7 @@ static int command(void)
     } /* if */
     check_empty(lptr);
     break;
-  case tINCLUDE:                /* #include directive */
+  case tpINCLUDE:                /* #include directive */
```
Southclaws: Why has this symbol gained a `p`?

ziggi: Name consistent with others
image

Y-Less:
This seems like a small but fundamental change internally.  How extensively has it been tested on scripts that push the compiler in unusual ways?  Basically, have you tried building YSI and its tests?

source/compiler/sc2.c
YashasSamaga: It mostly stands for: `t[oken]p[reprocessor]include`

YashasSamaga:
Same output in 3.10.9 and the compiler from this PR.

```
Tests: 177831, Fails: 15
    - Test:y_commands_removable2 (1)
    - Test:y_commands_removable (3)
    - Test:y_commands_lengths (6)
    - Test:y_commands_1 (5)
```

Original compiler:
```
*** Tests: 177831, Fails: 16
    - Test:y_commands_removable2 (1)
    - Test:y_commands_removable (3)
    - Test:y_commands_lengths (6)
    - Test:y_commands_1 (5)
    - Test:va_return_PassParameters (1)```

**Comment by Y-Less** [comment]
Well less fails than the original compiler is a good thing.

File: source/compiler/sc2.c [hunk 2]
```diff
@@ -2133,10 +2133,10 @@ SC_FUNC void lexinit(void)
   _lexnewline=FALSE;
 }
 
-char *sc_tokens[] = {
+char *sc_tokens[tlexLAST-tFIRST+1] = {
```
Y-Less: This seems to be the only place `tlexLAST` is used.  It wasn't in the original, doesn't fit with the naming scheme of the other meta-tokens, could be confused with `tLAST`, and isn't needed to define the array size.  So why it is used?

YashasSamaga: It was a futile attempt to link `sc_tokens` with the token list.

I'll remove it.

YashasSamaga:
Merge?

PR #397, opened 2018-12-31
More fixes related to suggestions

Daniel-Cortez:

This PR fixes the following bugs related to error suggestions:
* Functions aren't suggested before the point of their definition/implementation:
```Pawn
main()
{
    // The compiler is supposed to suggest "DoNothing" here, but it doesn't,
    // even though Pawn functions can be called before being declared/implemented.
    DoNothin(); // error 017: undefined symbol "DoNothin"
}

DoNothing(){}
```

* Failed assertion when `tagof` is used on a string or numeric literal:
```Pawn
main()
{
    // This gives an "assertion failed" message if the compiler is built
    // with assertions enabled
    return tagof(0);
}
```

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -508,13 +508,18 @@ enum {  /* identifier types */
   estAUTOMATON,
   estSTATE
 };
-enum {  /* symbol types */
-  essNONLABEL,  /* find symbols of any type but labels */
-  essVARCONST,  /* array, single variable or named constant */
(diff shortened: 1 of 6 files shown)


Y-Less:
Any objections?  @Zeex @Southclaws @YashasSamaga ?  Otherwise I'll merge it.

PR #421, opened 2019-05-04
A batch of improvements for the '__emit' operator

### LINKED ISSUE #346: new features for __emit

* [ ] Bug Report
* [x] Feature Request
* [ ] Question

**What happened**:
I have a few ideas for the `__emit` operator, but before implementing them and making a PR I guess it would be better to discuss them here first.

1. Add `__emit dump <identifier>`.
   This would allow the compiler to retrieve addresses of functions, variables etc. as constant values, which could be useful for initialization of constants and immutable arrays.
   Example:
```Pawn
static const functions[] =
{
    __emit dump func1,
    __emit dump func2,
    __emit dump func3
};
```
   Though I'm not sure if `dump` would be a proper pseudo-instruction name because in compiler assembly output `dump` means "put value(s) into the data section", not "retrieve symbol value/address".

2. Allow jumps to naked functions.
   Since the next release is going to have naked functions without prologue and epilogue code, maybe we should allow such functions as arguments for jump instructions?
```Pawn
#pragma naked
NakedFunction() { /* ... */ }
// ...
__emit jzer NakedFunction;
```

3. Issue an error/warning if data/stack offset is not a multiple of cell size.
```Pawn
__emit load.s.pri 5; // The offset is not a multiple of 4, so this is clearly an error.
```
   This would require addition of new error in `sc5.c` though (would "not a multiple of cell size" be a good description for it?)

4. Add a new error for invalid IDs in `lctrl`, `sctrl` and other instructions.
```Pawn
__emit lctrl -1;
```
   Currently this would make the compiler issue an "invalid range" error, which clearly isn't the right description. But again, I'm not sure how to phrase that error properly.

Special thanks to @VVWVV for giving me the idea about the "not a multiple of cell size" error.

**Environment**:

* Operating System
* Compiler version

### LINKED ISSUE #412: Crash if unused emit native
### Issue description:

Using `emit sysreq.c native;` without using the native elsewhere in the code will crash compiler.

```
native print(const str[]);

main() {
    new str[12];
    emit {
        push.adr str
        push.c 4
        sysreq.c print
        stack 8
    }
}
```

### Workspace Information:

* Compiler version: 3.10.9
* Command line arguments provided (or sampctl version): -;+ -)+
* Operating System: Windows 8.1

### LINKED ISSUE #413: Operator emit negative values
### Issue description:

Some emit instructions that should take negative integers don't. For example ADDR.pri -20, LOAD.S.pri -20, STOR.S.pri -20, ...

```
main() {
	emit STACK -20; // Allocate 5 cells
	emit ADDR.alt -20; // Get address of this array of 5 cells
        emit ZERO.pri;
	emit FILL 20; // Fill it with zeros
	emit STOR.S.pri -20; // Access first element
	emit STACK 20; // Free
}
```

### Workspace Information:

* Compiler version: 3.10.9
* Command line arguments provided (or sampctl version): -;+ -(+
* Operating System: Windows 8.1

Daniel-Cortez:

* Added 'universal' pseudo-opcodes, designed to be primarily used in macros.
  `load.u.pri/alt` - load value (may be an expression) into PRI/ALT.
  `stor.u.pri/alt` - store value in PRI/ALT into the operand.
  `addr.u.pri/alt` - load operand address into PRI/ALT.
  `push.u` - push value (may be an expression) into the stack.
  `push.adr.u` - obtain operand address and push it into the stack.
  `zero.u` - write value `0` into the operand.
  `inc.u` - increase the value stored in the operand.
  `dec.u` - decrease the value stored in the operand.

  Example:
```Pawn
#define getval(%0) __emit(load.u.pri %0)
#define addrof(%0) __emit(addr.u.pri %0)

const global_const = 0x1234;
new global_var = 0x5678;

main()
{
	const local_const = 0x4321;
	new local_var = 0x8765;
	static local_static_var = 0x9ABC;
	static local_static_array[2];

	printf("%08x", getval(global_const)); // 00001234
	printf("%08x", getval(global_var)); // 00005678
	printf("%08x", getval(local_const)); // 00004321
	printf("%08x", getval(local_var)); // 00008765
	printf("%08x", getval(local_static_var + global_const)); // 0000ACF0
	printf("%08x", addrof(global_var)); // 00000000
	printf("%08x", addrof(local_var)); // 00004094
	printf("%08x", addrof(local_static_var)); // 00000004
	printf("%08x", addrof(local_static_array[1])); // 0000000C
}
```

* The compiler now issues an error if the stack offset/data address isn't a multiple of cell size.
  NOTE: For this purpose I reused error 011 which previously wasn't used anywhere.
```Pawn
__emit load.s.pri 5; // error 011: stack offset/data address must be a multiple of cell size
```

* Fixed crash (#412) when referencing a previously unused native function.

* Removed index check for operands of `lctrl` and `sctrl`.

* Opcodes `stor.s.pri/alt`, `inc.s`, `dec.s`, and `zero.s` don't accept variables and arrays passed by reference anymore.

* Allowed negative offsets for arguments of type 'local variable' (#413).

* Added tests.
  NOTE: Since the compiler can only display up to 26 errors, I had to split the tests into 7 different *.pwn files (+ the extra one to check the generated bytecode).

* Fixed a lot of minor bugs discovered with the newly added tests.

Closes #346
Fixes #412
Fixes #413

* [x] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:


Daniel-Cortez:
There are a few things I'm still not sure about:
* Array handling is only done for pseudo-opcodes `load.u.pri/alt` and `push.u` - this is because for these opcodes I reused the standard expression parser (function `expression()` from `sc3.c`).
  For the other pseudo-instructions the operand is not an expression but an lvalue, so it's required to know its address, and I couldn't find any existing functionality for this task in `sc3.c`.
  As a temporary solution I used function `expression()` for those opcodes as well: if the user specifies a variable or a reference (`iVARIABLE`/`iREFERENCE`) the function returns the corresponding `symbol *`, and it's possible to know its address (`sym->addr`), but if the user specifies an array cell (`iARRAYCELL`; e.g. `arr[0]`), there seems to be no way to obtain its address.
  I have no idea how to implement array handling for those opcodes in a good way, without reinventing the wheel, so any help or suggestions for this would be appreciated.

* Currently instructions `casetbl` and `case` can be used in single-instruction `__emit` statements and in expressions, but it's easy to make the compiler generate invalid code:
```Pawn
__emit casetbl 2 lbl_default;
__emit case 0 lbl_case0;
new x = 0; // This line will become a part of the case table, which is obviously wrong
__emit case 1 lbl_case1;
```
  I think it would be better to only allow the use of those instructions in block mode, like this:
```Pawn
__emit
{
	casetbl 2 :lbl_default
	case 0 :lbl_case0
	case 1 :lbl_case1
lbl_default:
	// ...
lbl_case0:
	// ...
lbl_case1:
	// ...
}
```
  That way it would be impossible to put any other code inbetween case table entries, but this would require adding a new error and I'm not sure how to phrase such error (not a native English speaker) - any help/suggestions on how to properly phrase it would be nice.

  If and when this is done, it would be possible to improve control over the `case` entries:
  * Check if the number of entries matches the number specified in the `casetbl` instruction.
  * Make sure the entries in the table are sorted by value (in the Implementer's Guide it's said they must be sorted so the Abstract Machine could use binary search on them).

Daniel-Cortez:
Ok, I reimplemented the pseudo-opcodes, now all of them can handle array access.
The only exception is that array characters (e.g. `array{n}`) are not allowed in `push.u.adr`: I made it that way, because in normal non-`__emit` Pawn array characters can't be passed to functions by reference.

I still need help with formulating the error messages in order to improve control over case table entries made with `__emit` (see the previous post).
But if that feature isn't needed much, then I suppose this PR should be ready to be merged.

Zeex:
Looks good 👍  though I'm not well versed in the new `__emit` stuff

PR #423, opened 2019-06-06
Misc. minor fixes

### LINKED ISSUE #425: Incorrect fatal error when compact encoding is ineffective
### Issue description:
The compiler issues an incorrect error for the following code:
```Pawn
main()
{
    // NOTE: In compact encoding '0xAABBCCDD' can't be compressed well
    // and takes 5 bytes instead of 4, thus making the compression ineffective.
    static const a[32] = { 0xAABBCCDD, ... };
    return a[0];
}
```
Since the compression shouldn't be effective for this code, one would expect the compiler printing the following message:
```
warning 232: output file is written, but with compact encoding disabled
```
But instead it issues a fatal error (and aborts compilation as a result):
```
fatal error 106: compiled script exceeds the maximum memory size (1521262808 bytes)
```
And instead of `1521262808` there might be any other (seemingly random) value.

See above.

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.9
* Operating System:

Daniel-Cortez:

Fixes for various minor bugs that weren't worth making a separate PR for each of them:

* Fixed invalid formula for calculating array size (in characters) in function `hier1()`.
  The cell size in bits was hardcoded as `32` (should be `sizeof(cell)*8`), which could lead to generation of invalid code for array boundary checks if the Pawn toolkit was configured with 16- or 64-bit cell size.
  Of course, this bug doesn't affect SA-MP (which is 32-bit), but I think it's still good to have this fix here.

* Fixed invalid type cast in `funcstub()`.
  See https://github.com/pawn-lang/compiler/issues/296#issuecomment-450643026.

* Fixed OOB array access in `encode_cell()`.
  This one seems to originate from a commit that I made a year and a half ago (sorry!)

* Fixed memory leaks in `plungequalifiedfile()`.
  Special thanks goes to @YashasSamaga for [reporting the bug, as well as providing a fix](https://github.com/pawn-lang/compiler/issues/296#issuecomment-449903004), though I slightly changed that fix in the following ways:
  * Moved `free(path)` before `error(103)`, because `path` won't be freed otherwise. This is because on fatal errors (1xx) `error()` returns (via `longjmp()`) directly to an error buffer set in `pc_compile`, so the code after such `error()` call is never executed.
  * For the same reason I removed two `return`s, as they are pretty much useless after `error(<fatal error code>)`.

* Removed `NULL` pointer checks before the uses of function `free()`.
  [`free()` does nothing if the pointer is `NULL`](https://en.cppreference.com/w/c/memory/free), so any `NULL` checks before the use of `free()` are redundant.

* Fixed invalid uses of `realloc()` in several places.
  The old pointer could have got leaked if the function failed to resize the old memory block or allocate a new one.

* Simplified the code in function `pushstk()` by replacing a group of calls to `malloc()`/`memcpy()`/`free()` with one `realloc()`.

* Removed function `do_switch()` (`sc6.c`).
  Apparently this function was a duplicate of `do_jump()`; it didn't implement any extra checks, so its only difference from `do_jump()` was in its name.

* Simplified string emptiness checks.
  Using `strlen(<string>)==0` to identify empty strings is inefficient because if the string is not empty `strlen()` traverses through the whole string to get its length.

* Fixed potential `NULL` pointer dereferences in functions `sc_path()` and `setconfig()`.

* Fixed incorrect error message when compact encoding is ineffective (#425).

* Other minor adjustments.

Fixes #425

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/memfile.c
+++ b/source/compiler/memfile.c
@@ -90,14 +90,11 @@ int memfile_write(memfile_t *mf, const void *buffer, size_t size)
 	assert(buffer != NULL);
 	if (mf->offs + size > mf->size)
 	{
-		char *orgbase = mf->base; /* save, in case realloc() fails */
 		size_t newsize = (mf->size + size) * 2;
-		mf->base = (char *)realloc(mf->base, newsize);
-		if (!mf->base)
-		{
-			mf->base = orgbase;     /* restore old pointer to avoid a memory leak */
+    char *newbase = (char *)realloc(mf->base, newsize);
+		if (!newbase)
 			return 0;
-		}
+		mf->base = newbase;
 		mf->size = newsize;
 	}
 	memcpy(mf->base + mf->offs, buffer, size);
--- a/source/compiler/memfile.c
+++ b/source/compiler/memfile.c
@@ -90,12 +90,12 @@ int memfile_write(memfile_t *mf, const void *buffer, size_t size)
 	assert(buffer != NULL);
 	if (mf->offs + size > mf->size)
 	{
-	  char *orgbase = mf->base; /* save, in case realloc() fails */
+		char *orgbase = mf->base; /* save, in case realloc() fails */
 		size_t newsize = (mf->size + size) * 2;
 		mf->base = (char *)realloc(mf->base, newsize);
 		if (!mf->base)
 		{
-		  mf->base = orgbase;     /* restore old pointer to avoid a memory leak */
+			mf->base = orgbase;     /* restore old pointer to avoid a memory leak */
 			return 0;
 		}
 		mf->size = newsize;
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -59,6 +59,10 @@
   #define SC_FASTCALL
 #endif
 
+#if !defined strempty
+  #define strempty(str) ((str)[0]=='\0')
+#endif
+
 /* Note: the "cell" and "ucell" types are defined in AMX.H */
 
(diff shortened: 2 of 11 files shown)


Y-Less:
Looks fine to me, but only from a visual readthrough.

>       * For the same reason I removed two `return`s, as they are pretty much useless after `error(<fatal error code>)`.
> 

I couldn't find those returns.  Did you change this part?  I'd suggest not removing returns after fatal errors, since the c compiler doesn't know that `error` might not return (it could be marked as `noreturn`, but it does in some circumstances).  Thus the return will cut off certain code paths and allow the compiler compiler to do more when compiling the compiler.

Daniel-Cortez:
> I couldn't find those returns. Did you change this part?

No, I meant that I removed them from the code @YashasSamaga added in their original fix.

> I'd suggest not removing returns after fatal errors, since the c compiler doesn't know that error might not return (it could be marked as noreturn, but it does in some circumstances). Thus the return will cut off certain code paths and allow the compiler compiler to do more when compiling the compiler.

Makes sense, but I removed those returns not only because they would be unreachable, but also to be more in-line with the existing code (in all the other code there are usually no returns after fatal errors).

Y-Less:
Oh I see, then fair enough.

Daniel-Cortez:
@Zeex Done.

Zeex:
Thanks, looks good

PR #427, opened 2019-06-16
Ignore `#if`d out unknown directives.

### LINKED ISSUE #426: Ignoring unknown directives.
In short, this should be allowed:

```pawn
#if 0
	#unknown_directive
#endif
```

Currently that gives an error, when it should be skipped.  Implementing this is trivially easy - I already did it:

https://github.com/Y-Less/compiler/commit/267c8f8ffdb250ba855aefda6b52f84d5d8fac0a

But there's one HUGE issue - what happens if that unknown and ignored directive from a future compiler version actually ends or somehow changes the `#if`?  E.g:

```pawn
#if 0
	printf("never");
#fi
```

An older compiler that sees that will just ignore `#fi`, not end the `#if`, and probably reach the end of the file without doing anything.  There are two solutions:

1. Ignore this problem.  If you want to make your code backwards-compatible just don't rely on older compilers correctly ignoring (or not) important directives.

2. Enumerate all possible future control-flow directives there ever may be, this could be, but is not limited to:

```pawn
#if
#else
#elseif
#elif
#ifdef
#elifdef
#elseifdef
#endif
#fi
#el
#ifndef
#elseifndef
#elifndef
```

Will we ever want `#iff`?  `#switch`, `#esac` etc?  We can't possibly know.

Thus I think option 1 is fine - just don't be an idiot!

Y-Less:
Fixes #426 

Makes the compiler slightly more future-proof by allowing us to do something like:

```pawn
#if __Pawn >= 0x03AB
    #message You're on a nice new compiler.
#endif
```

Currently to do that you need to do:

```pawnc
#message You're on a nice new compiler.
```

```pawn
#if __Pawn >= 0x03AB
    #include "message.inc"
#endif
```

Or you get an error that `#message` (or anything else) is an unknown directive if the compiler doesn't recognise it, even if it is disabled.

The only downside to this is that it makes it slightly harder to add new `#if`-like directives:

```pawn
#if __Pawn >= 0x03AB
    #message You're on a nice new compiler.
#elif SHOW_ERROR
    #error You've got an error.
#endif
```

A compiler that fails the first check AND doesn't recognise `#elif` will simply skip it and go all the way to `#endif`.  It will not evaluate it as `#elseif` and potentially run the code within.  It should somehow recognise `#elif` as a `#if`-like directive it just doesn't know and give an error.  That would require some sort of hard convention we don't yet have.  That, or we just define all possible ones now (even if they aren't implemented):

```pawn
#if
#else
#elseif
#elif
#ifdef
#elifdef
#elseifdef
#endif
#fi
#el
#ifndef
#elseifndef
#elifndef
```

Can't think of any more, but that doesn't mean there aren't any (does `iff` make any sense here?)

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1558,8 +1558,10 @@ static int command(void)
     } /* if */
     break;
   default:
-    error(31);          /* unknown compiler directive */
-    ret=SKIPPING ? CMD_CONDFALSE : CMD_NONE;  /* process as normal line */
+    if (!SKIPPING) {
+      error(31);          /* unknown compiler directive */
+      ret=CMD_NONE;  /* process as normal line */
+    }
   } /* switch */
   return ret;
 }


Zeex:
Merged, but there is one issue - unknown directives cause an assertion failure (it was like this in earlier versions too, not due to this change). I'll push a separate fix for that.

PR #432, opened 2019-06-22
Misc. minor fixes, code cleanup

Daniel-Cortez:

This PR contains several minor fixes that aren't worth of making a separate PR for each of them.

* Moved macro `arraysize` from `amx.h` to `sc.h`, as previously discussed in https://github.com/pawn-lang/compiler/pull/397#commitcomment-33954084, and made use of it where appropriate.
  Originally I wanted to also rename it to `sizearray`, as this is how this macro was named in Pawn 3.3, but I'm not sure if we should do the same here. After all, this fork already has a lot of distinctions and improvements over the original codebase, so I believe one macro won't really make a difference (also the name `arraysize` simply sounds better, as in `sizearray` the words `array` and `size` are put in reverse order).
Another difference from the original code is that I also made a special version of `arraysize` for GCC and Clang. One of my main concerns about using that macro was that it's easy to accidentally use it on pointers and the compiler won't prevent the coder from doing so, so I added an 'extended' version of that macro that uses GCC-specific language extensions to reject pointers.

* Fixed potential undefined behavior in 'alignframe()' (`sc4.c`).

* Fixed the use of wrong floating-point type in a sanity check in function `ftoi()` (`sc2.c`).
  Also added a few static assertions to make sure type punning from `float`/`double` to `cell` is safe.

* Removed various leftover stuff:
  * In function `stgopt()` (`sc7.c`) assertion `assert(sequences!=NULL)` is redundant as `sequences` is an array and thus can't be `NULL` (it was a pointer in the original Pawn 3.2 code though).
  * Functions `phopt_init()` and `phopt_cleanup()` were removed from `sc7.c`, but their declarations in `sc.h` weren't.
  * Same with functions `find_alias()` and `get_subst()`, except they were removed in the original Pawn source (before Pawn 3.0).

* Other minor adjustments.

Fixes #

* [x] A Bug Fix
* [ ] A New Feature
* [x] Other

**Additional Documentation**:


Daniel-Cortez:
Removed whitespace changes, as they could mess up the other PRs. Not sure if there's anything else to do in the context of this PR.

Daniel-Cortez:
Is there anything that prevents this PR from being merged?

Daniel-Cortez:
Excluded the commit that fixes a typo in the error message code (`7` => `47`) from this PR, as I moved it into a separate PR (#499).

PR #437, opened 2019-07-05
Fix pseudo-opcodes being able to modify constant variables/arrays

Daniel-Cortez:

Fixes a small oversight in that the newly added pseudo-opcodes (`stor.u.pri/alt`, `zero.u`, `inc.u`, `dec.u`) are able to modify constant variables.
Example:
```Pawn
new const x = 1;
__emit zero.u x; // Before the fix: no errors/warnings; 'x' is zeroed despite being 'const'
                 // After: error 022: must be lvalue (non-constant)
```
These pseudo-opcodes are designed to be used in macros, so they're supposed to stick to the normal Pawn rules as closely as possible, thus they shouldn't be able to modify constant variables/arrays.

-

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:


Y-Less:
Normal opcodes can modify constants.  I use that in a few places for some unobservable mutations:

When printing part of a string, fixes.inc null-terminates it, then later restores the character, just to avoid copying the whole string.  But the function can take constant strings, because they don't end differently to how they started.

Also, YSI uses this trick for `final` variables.  They are `const` so that they can't be modified, but still need to be written to once to get their value.

Daniel-Cortez:
Normal opcodes will still be able to modify constants, this change only applies to macro-opcodes. For example, this may prevent the user from mistakenly modifying a constant variable/array with a macro.

Y-Less:
Fair enough.

source/compiler/sc1.c
```diff
@@ -6063,8 +6063,9 @@ static regid SC_FASTCALL emit_findreg(char *opname)
  * Looks for an lvalue and generates code to get cell address in PRI
  * if the lvalue is an array element (iARRAYCELL or iARRAYCHAR).
  */
-static int SC_FASTCALL emit_getlval(int *identptr,emit_outval *p,int *islocal, regid reg,
-                                    int allow_char, int store_pri,int store_alt,int *ispushed)
+static int SC_FASTCALL emit_getlval(int *identptr,emit_outval *p,int *islocal,
+                                    regid reg,int allow_char,int allow_const,
+                                    int store_pri,int store_alt,int *ispushed)
```
Y-Less: Why this change?

Issue #267, opened 2018-01-19
__attribute__((attr-list)) or [[attr-list]]


YashasSamaga:
`#pragma` is a preprocessor directive; hence, apart from being ugly, it puts limitations on its usage. For example, pragma cannot be used inside substitution pattern of a `#define` (`of a  ??`; what is the correct technical term?).

If something like the `__attribute__(())` in GCC or like `[[]]` in C++17 (or `__declspec`  in Microsoft; not symmetric and not as readable as the other two in my opinion) is added, this would get rid of those limitations and also improve readability of the code.

# Design ideas

## Basic Examples:
Some the attributes used in this section do not exist in the current compiler. I have used them to just present the possible syntax so that viewers can get an idea how it would look when used in code (some cases such as statement attributes do not have any options that could be used in the current version of the compiler but they could be added in the future so let's consider those as well).

 - function attributes
`stock __attribute__((naked)) myfunc () { }` 
`stock myfunc () __attribute__((naked)) { }` 
`[[naked]] stock myfunc () { }` 
`stock myfunc () [[naked]] { }` 

 - function argument attributes
`stock myfunc(__attribute__((unused)) myarg, normal_arg) { } `
`stock myfunc([[unused]] myarg, normal_arg) { }`
`stock myfunc(myarg __attribute__((unused)) , normal_arg) { } `
`stock myfunc(myarg [[unused]] , normal_arg) { }`

 - variable attributes
`new __attribute__((unused)) var;`
`new [[unused]] var;`
`new var __attribute__((unused));`
`new var [[unused]] ;`

 - label attributes
`__attribute__((omit-defaults)) mylabel:` (potential solution for a part of #266 )
`[[omit-defaults]] mylabel:` 
`mylabel: __attribute__((omit-default));`
`mylabel: [[omit-defaults]];` 

 - enumerator attributes
 ```
 enum {
       __attribute__((deprecated)) score,
       [[deprecated]] money,
       points __attribute__((deprecated)),
       something [[depreacted]]
 };

enum __attribute__((unused)) { }
enum [[unused]] { }
 ```
 
- statement attributes
`__attribute__((nodestruct)) return 0;`
`[[nodestruct]] return 0;`

## Empty attributes:
`stock __attribute__(()) myfunc() { }`
`new __attribute__(()) myvar;`

This might provide more flexibility when attributes are a part of substitution pattern.

## Multiple attributes:

[...]

PR #446, opened 2019-09-15
Operator '__fallthrough'

Daniel-Cortez:

This PR implements operator `__fallthrough`, which, as the name suggests, can be used to make "fall-through" switch cases.
Example:
```Pawn
SendClientMessage(playerid, -1, "Available admin commands:");
switch (player_info[playerid][pAdminLvl])
{
    case 3:
    {
        SendClientMessage(playerid, -1, "Level 3: /makeadmin [playerid] [level]");
        __fallthrough; // Show commands for the previous admin levels as well
    }
    case 2:
    {
        SendClientMessage(playerid, -1, "Level 2: /tphere [playerid], /abroadcast [text]");
        __fallthrough;
    }
    case 1:
    {
        SendClientMessage(playerid, -1, "Level 1: /kick [playerid] [reason], /ban [playerid] [reason]");
        __fallthrough; // Silently ignored, as this is the last case
    }
}
```

Fixes #

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:


Y-Less:
I'm not convinced this is needed,  Pawn cases were explicitly designed to counter the common issues with C cases from fallthrough.  While explicit is better than implicit, you can already do it with `goto`.

Daniel-Cortez:
> Pawn cases were explicitly designed to counter the common issues with C cases from fallthrough.

As far as I can understand, `__fallthrough` shouldn't cause any of those issues since its used in an explicit way.

> While explicit is better than implicit, you can already do it with goto.

`goto` requires the user to define a label in the next case and generates extra code (one `jump` and a pair of `lctrl`/`sctrl` to readjust the stack pointer), both of which, IMHO, are excessive for a simple fallthrough case. Operator `__fallthrough` doesn't do any of that.

YashasSamaga:
I think we should have a generic system. In C++, this would be done using `[[fallthrough]]` where `[[]]` is called attribute specifier sequence.

There was a discussion earlier about such a thing #267.

Y-Less:
You're talking about adding a whole new keyword (which should only be done extremely sparingly) to save two instructions in a rare corner-case.  That's why I don't think it is worth it.

Y-Less:
The real response should be improving the behaviour of `goto` to not reset the stack when it doesn't need to.  Or just adjust it by however much is required to change between two variable scopes.  This solution does also have stack adjusting ops, just not in any of the given tests, because none of those have case-local variables.

Daniel-Cortez:
> You're talking about adding a whole new keyword (which should only be done extremely sparingly) to save two instructions in a rare corner-case. That's why I don't think it is worth it.

Then would it be OK to reimplement this feature as `#pragma fallthrough`? If I understand it correctly, `#pragma` **is** pretty much designed for rare corner cases.

>The real response should be improving the behaviour of goto to not reset the stack when it doesn't need to. Or just adjust it by however much is required to change between two variable scopes.

I think this would require implementing an ability to reparse functions, because the compiler can't know in advance in which scope the label is if it's used in `goto` before the point of its definition. This would also open a way to fix other related problems (e.g. destructors for local variables not being called when branching out of their scope with `goto`).

PR #453, opened 2019-09-30
Fix assertion failure whence the script contains a single state variable

### LINKED ISSUE #429: Assertion failure with a state variable
### Issue description:
Compiling the following code
```Pawn
new var1 <STATE_1>;
main(){}
```
would lead to assertion failure
```
Assertion failed: sym==NULL || sym->states==NULL && sc_curstates==0 || sym->states!=NULL && sym->next!=NULL && sym->states->first->index==sc_curstates, file sc1.c, line 2080
```
How to reproduce this error:
* Build the compiler with assertions enabled.
* Make sure there's no `default.inc`, or if it's present, it shouldn't `#include` anything else (so there wouldn't be other symbols after the state variable in the global symbols list).
* Try to compile the code listed above.

The condition that fails in that assertion is
```C
sym->states!=NULL && sym->next!=NULL && sym->states->first->index==sc_curstates
```
and it fails because `sym->next` is `NULL` (and it's supposed to be `NULL` since `var1` is the only variable in the script and it happens to be the last symbol in the global symbols table).

Adding another variable after `var1`
```Pawn
new var1 <STATE_1>;
new var2;
main(){}
```
fixes the error.

Do we really need the `sym->next!=NULL` check in that assertion?

See above.

### Workspace Information:

* Compiler version:
* Operating System:

Daniel-Cortez:

See #429 for the explanation on how to reproduce the bug.
https://github.com/pawn-lang/compiler/blob/2f642899f1eb283ea5c40143b98112eb9fc19568/source/compiler/sc1.c#L2079-L2081
It seems that the original author of this code meant to write `sym->states->next!=NULL` instead of `sym->next!=NULL` (note that in this fork the same should be written as `sym->states->first!=NULL` because we have `constvalue_root`). This is confirmed by the fact that `sym->states->first` (or `sym->states->next` in the original code) gets dereferenced right after that subexpression (`sym->states->first->index==sc_curstates`).

Fixes #429

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -2078,7 +2078,7 @@ static void declglb(char *firstname,int firsttag,int fpublic,int fstatic,int fst
      */
     assert(sym==NULL
            || sym->states==NULL && sc_curstates==0
-           || sym->states!=NULL && sym->next!=NULL && sym->states->first->index==sc_curstates);
+           || sym->states!=NULL && sym->states->first!=NULL && sym->states->first->index==sc_curstates);
     /* a state variable may only have a single id in its list (so either this
      * variable has no states, or it has a single list)
      */


Southclaws:
Thanks for the fix!

It would be nice to have a test harness around this just to prove correctness automatically and prevent regressions of the same bug.

Daniel-Cortez:
@Southclaws I'd do that, but this bug is very specific - in order to reproduce it I had to either a) remove `default.inc` in the `include` directory, or b) comment out the `#include <core>` and `#include <console>` lines inside that file (so the global symbol table wouldn't be populated by the symbols from `core.inc` and `console.inc` included by default and would only contain the state variable from the test code, otherwise the assertion wouldn't trigger). I don't think it would be alright to do either of those in this repo (or would it?)

PR #455, opened 2019-09-30
#pragma fallthrough

Daniel-Cortez:

This PR implements `#pragma fallthrough` that allows to make fall-through switch cases.
Example:
```Pawn
SendClientMessage(playerid, -1, "Available admin commands:");
switch (player_info[playerid][pAdminLvl])
{
    case 3:
    {
        SendClientMessage(playerid, -1, "Level 3: /makeadmin [playerid] [level]");
        #pragma fallthrough // Show commands for the previous admin levels as well
    }
    case 2:
    {
        #pragma fallthrough // Make this case fallthrough as well
        SendClientMessage(playerid, -1, "Level 2: /tphere [playerid], /abroadcast [text]");
    }
    case 1:
    {
        SendClientMessage(playerid, -1, "Level 1: /kick [playerid] [reason], /ban [playerid] [reason]");
        #pragma fallthrough // Silently ignored, as this is the last case
    }
}
```
Technically this is a rework of #446 (operator `__fallthrough`), which I decided to revise after discussion about wasting a keyword for what's expected to be a rarely used feature.

Fixes #

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -920,6 +920,8 @@ SC_VDECL int pc_memflags;     /* special flags for the stack/heap usage */
 SC_VDECL int pc_naked;        /* if true mark following function as naked */
 SC_VDECL int pc_compat;       /* running in compatibility mode? */
 SC_VDECL int pc_recursion;    /* enable detailed recursion report? */
+SC_VDECL int pc_fallthrough;  /* is current switch case fallthrough? */
+SC_VDECL int pc_inswitch;     /* are we currently inside a switch case? */
 
 SC_VDECL constvalue_root sc_automaton_tab; /* automaton table */
 SC_VDECL constvalue_root sc_state_tab;     /* state table */
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -903,6 +903,8 @@ static void resetglobals(void)
   sc_curstates=0;
   pc_memflags=0;
   pc_naked=FALSE;
+  pc_fallthrough=FALSE;
+  pc_inswitch=FALSE;
   emit_flags=0;
   emit_stgbuf_idx=-1;
 }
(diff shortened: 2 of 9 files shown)


Y-Less:
The gist of the discussion in #446 was "just use goto".  The main argument against that being "but that wastes two instructions".  So previously the cost/reward was a new keyword vs 2 instructions, which was IMHO way too high.  Now it is a new pragma vs 2 instructions, which granted is a much lower cost/reward ratio as pragmas are already somewhat gated as advanced features, but even then I still don't think it is worth it.  In the grand scheme of things, with the way people code and the way the compiler (doesn't) optimise, there are far better ways to save two instructions should it ever prove necessary.

Southclaws:
I think another issue here (that I should have thought about a long time ago!) is the fact you're taking the time to write these features (which, of course is appreciated!) that implement new language features that obviously warrant discussion beforehand.

So, in response I'm going to propose a section in the readme/contributor guidelines that new features and complex bug fixes be discussed via issues before being worked on, just to save time and prevent work from being wasted.

Daniel-Cortez:
> The gist of the discussion in #446 was "just use goto".

You complained about wasting an operator name on that feature, so I asked if it would be alright to reimplement it as a `#pragma`, but didn't get an answer (even though you usually answered the other messages in that issue in a few hours). Probably was my fault for interpreting that as a silent agreement.

> just use goto

(I'm probably repeating myself, but whatever.)
Doing that takes two lines of code (one for `goto <label>` and the other one for the label declaration) in two different switch cases, while with `#pragma fallthrough` it's only one line in one place and the user doesn't have to think on how to name the label.

> The main argument against that being "but that wastes two instructions".

Actually 4, but these times that's probably nothing anyway. In any case, my main motivation for this feature is to give the user a more convenient way to use fallthrough switch cases when they would need it.

PR #456, opened 2019-09-30
enum continuation

Daniel-Cortez:

This PR adds an ability to continue previously defined enumerations.
Example:
```Pawn
enum dialogid // note that the tag 'dialogid' is weak (starts with a lowercase letter)
{             // so tag override isn't required when using the constants from this enum
    DIALOG_ID_NONE = 0
};

continue enum dialogid // continue the enumeration
{
    DIALOGID_REGISTER, // 1
    DIALOGID_LOGIN     // 2
};

enum continue dialogid // 'enum continue' is also valid
{
    DIALOGID_HELP,     // 3
    DIALOGID_RULES     // 4
};
```
This could be used in different scenarios. The most obvious one, as shown above, is to define SA-MP dialog IDs in different source files for better modularity.

Fixes #

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

* Specifying a enum name after `continue enum`/`enum continue` is mandatory (or else how would the compiler know which enum to continue?)
```Pawn
enum continue // error (no enum name)
{
    // ...
};
```
* The first enum in the continuation chain can also be defined via `continue enum`/`enum continue`.
```Pawn
#include <a_samp>

continue enum dialogid // enum 'dialogid' wasn't previously defined, so the compiler
{                      // would simply create a new one and ignore the 'continue' keyword
    DIALOG_ID_NONE // 0
};
```
* Enum continuation can't be used inside functions.
* `continue enum`/`enum continue` can't be combined with static enumerations ( `static enum`/`enum static`).

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -93,7 +93,7 @@ static void declglb(char *firstname,int firsttag,int fpublic,int fstatic,
                     int stock,int fconst);
 static int declloc(int fstatic);
 static void decl_const(int table);
-static void decl_enum(int table,int fstatic);
+static void decl_enum(int table,int fstatic,int fcontinue);
 static cell needsub(int *tag,constvalue_root **enumroot);
 static void initials(int ident,int tag,cell *size,int dim[],int numdim,
                      constvalue_root *enumroot);
@@ -1707,7 +1707,7 @@ static void parse(void)
       break;
     case tSTATIC:
       if (matchtoken(tENUM)) {
-        decl_enum(sGLOBAL,TRUE);
+        decl_enum(sGLOBAL,TRUE,FALSE);
       } else {
         /* This can be a static function or a static global variable; we know
          * which of the two as soon as we have parsed up to the point where an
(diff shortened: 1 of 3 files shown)


Y-Less:
This is going the same way as #455 and #446 - extending the language to address insanely niche issues that already have perfectly good solutions (https://forum.sa-mp.com/showpost.php?p=4045875&postcount=702).  Yes the solutions may take an extra line of code, but it's an extra line of existing code instead of a whole new language feature raising the complexity of the language itself and the barrier to entry.

Plus your only example isn't even the best way to do dialogs (its been widely deprecated for years).  Can you provide any more cases where this would be useful, and not solve easily in any other way?

Also, what's the result of this:

```pawn
enum test
{
    A,
    B,
    C,
}

new arr[test];

continue enum test
{
    D,
    E,
}
````

?

Most new features are discussed in issues first, and those seen as worthwhile go to PR, not the other way around.

Y-Less:
Side note:  If you'd waited a day to make these PRs they would have all counted for Hacktoberfest.  You could have got a free t-shirt.

Issue #438, opened 2019-07-07
Old-style and new-style function declarations


Daniel-Cortez:
### Issue description:
New-style function declarations (starting with the `forward` keyword) after the function is implemented are OK, but apparently the old-style ones aren't:
```Pawn
public TestFunc();
public TestFunc(){}


forward TestFunc(); // no error
forward public TestFunc(); // no error
public TestFunc(); // error 021: symbol already defined: "TestFunc"
```
Is this a bug? If not, then maybe we should allow old-style declarations after the function is defined too?

### Minimal complete verifiable example (MCVE):
See above.

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.9
* Command line arguments provided (or sampctl version): -
* Operating System: -

Y-Less:
I don't think old-style anything missing something that new-style has is a bug.  There's no point updating old code to work the same as new code when we have the new code.

Daniel-Cortez:
The fact that the compiler calls old-style function declarations "old-style" doesn't really make them "old" or "outdated" as your answer implies. Their use may be uncommon among the SA-MP community, but there's not even a single recommendation against using them in the official documentation; the Language Guide rather refers to them as "an alternative way to forwardly declare a function". In fact, this syntax is even kept in Pawn 4.0.

I really hope you can reconsider your opinion on allowing to use such declarations after the implementation. This is just an alternative way to declare functions, not an outdated syntax.

PR #457, opened 2019-09-30
"static public" functions and variables

### LINKED ISSUE #306: Public functions with the same name

* [x] Bug Report
* [ ] Feature Request
* [ ] Question

**What happened**:
The compiler creates public functions with the same name.

Normally it issues a "symbol already defined" error, but in the example below it doesn't.

```Pawn
// myinc.inc
static @SomeFunc();
static @SomeFunc()
{
	emit const.pri 0;
}
```
```Pawn
// main.p
static @SomeFunc();
static @SomeFunc()
{
	emit const.pri 1;
}

main(){}
```
As a result, the compiler creates two entries named `@SomeFunc` in the public functions table, which clearly isn't right.

I believe this can be solved in two ways:
* If a function is defined with a `static` specifier and its name starts with `@`, check if there are other functions with the same name in different scopes and issue an error if so.

OR
* Do now allow to define `static` functions with name starting with `@` (the compiler already forbids a combination of `static` and `public`).

Related discussions:
http://pro-pawn.ru/showthread.php?16229
http://forum.sa-mp.com/showthread.php?t=654148

**Environment**:

* Operating System
* Compiler version

Daniel-Cortez:

This PR does the following:
* Allows to declare and define "static public" variables and functions that have visibility scope reduced to the source file they're defined in, but can be accessed from the VM as any other public functions or variables (as suggested in https://github.com/pawn-lang/compiler/issues/306#issuecomment-397771620).
```Pawn
static public OnSomethingHappen();
static public OnSomethingHappen()
{
    // ...
}
```

* Adds an extra check when creating a function/variable whose name starts with symbol `@` to make sure another function/variable with the same name wasn't declared with a `static` classifier in another source file (see #306).
```Pawn
// myinc.inc
static @SomeFunc();
static @SomeFunc()
{
    // ...
}
```
```Pawn
// main.pwn
#include "myinc.inc"

static @SomeFunc();
static @SomeFunc() // error (static public function already defined in another file)
{
    // ...
}
```

* Allows to re-declare functions after their use with old-style prototypes (for some reason this is already possible for new-style prototypes, but not for the old-style ones) (see #438).
```Pawn
// Before:
TestFunc(){}

forward TestFunc(); // no error
TestFunc(); // error 021: symbol already defined: "TestFunc"
```
```Pawn
// After:
TestFunc(){}

forward TestFunc(); // no error
TestFunc(); // no error
```
This also allows to specify the function state in old-style function prototypes (even though the compiler would print a warning about state specifications being ignored in function prototypes, which is what it already does for new-style declarations).

Fixes #306, #438

* [x] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -246,6 +246,7 @@ typedef struct s_symbol {
 #define sSTATIC   2     /* global life, local scope */
 
 #define sSTATEVAR  3    /* criterion to find variables (sSTATEVAR implies a global variable) */
+#define sGLOBALALL 4    /* criterion to find global symbols in all source files */
 
 typedef struct s_value {
   symbol *sym;          /* symbol in symbol table, NULL for (constant) expression */
@@ -643,7 +644,7 @@ long pc_lengthbin(void *handle); /* return the length of the file */
 
 /* function prototypes in SC1.C */
 SC_FUNC void set_extension(char *filename,char *extension,int force);
-SC_FUNC symbol *fetchfunc(char *name,int tag);
+SC_FUNC symbol *fetchfunc(char *name,int tag,int fpublic);
 SC_FUNC char *operator_symname(char *symname,char *opername,int tag1,int tag2,int numtags,int resulttag);
 SC_FUNC void check_index_tagmismatch(char *symname,int expectedtag,int actualtag,int allowcoerce,int errline);
 SC_FUNC void check_tagmismatch(int formaltag,int actualtag,int allowcoerce,int errline);
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -1680,12 +1680,6 @@ static int getclassspec(int initialtok,int *fpublic,int *fstatic,int *fstock,int
     } /* switch */
   } while (tok && err==0);
 
-  /* extra checks */
-  if (*fstatic && *fpublic) {
-    err=42;              /* invalid combination of class specifiers */
-    *fstatic=*fpublic=FALSE;
-  } /* if */
-
   if (err)
     error(err);
   return err==0;
@@ -1729,10 +1723,8 @@ static void parse(void)
          * deciding it was a declaration of a static variable after all, we have
          * to store the symbol name and tag.
          */
-        if (getclassspec(tok,&fpublic,&fstatic,&fstock,&fconst)) {
-          assert(!fpublic);
+        if (getclassspec(tok,&fpublic,&fstatic,&fstock,&fconst))
           declfuncvar(fpublic,fstatic,fstock,fconst);
-        } /* if */
       } /* if */
       break;
     case tCONST:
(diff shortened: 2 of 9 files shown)


Y-Less:
Nice, thanks.  I'll test this later tonight.

Y-Less:
Nice, thanks.  I'll test this later tonight.

Daniel-Cortez:
Updated to resolve merge conflicts.

Y-Less:
I'm a bit concerned about this as well, mostly due to the warnings.  You get a "symbol already defined" warning when the same `public static` is in two different files, but this is the opposite of the semantics of `static`.  So someone seeing that warning might reasonably complain that it isn't defined in the current compilation unit, so shouldn't be seen for the purposes of symbol resolution.  So the warning is using some hybrid of two totally different sets of scoping rules (as is all of this PR - the function is static, unless you call it a different way).

I think the underlying issue is what is the motivating use-case for this?  A function that should be called from the VM, but also called from the current file.  Most things called from the VM are callbacks, not normal functions; things called from user code are the opposite.  Are there any good examples of a function that legitimately needs to be both, and where this is by far the best solution?

Daniel-Cortez:
> Are there any good examples of a function that legitimately needs to be both, and where this is by far the best solution?

Don't want to sound rude, but I thought you know where this can be used, as it was you who initially [suggested this](https://github.com/pawn-lang/compiler/issues/306#issuecomment-397771620):
> I think static public should be allowed, since sometimes you do want to restrict being able to directly call the function to the module. Obviously this would require checking for two publics with the same name.

Daniel-Cortez:
Also, if you're going to close this PR, would you mind if I moved this part

> * Allows to re-declare functions after their use with old-style prototypes (for some reason this is already possible for new-style prototypes, but not for the old-style ones) (see #438).
>
> ```Pawn
> // Before:
> TestFunc(){}
> 
> forward TestFunc(); // no error
> TestFunc(); // error 021: symbol already defined: "TestFunc"
> ```
> 
> ```Pawn
> // After:
> TestFunc(){}
> 
> forward TestFunc(); // no error
> TestFunc(); // no error
> ```
> 
> This also allows to specify the function state in old-style function prototypes (even though the compiler would print a warning about state specifications being ignored in function prototypes, which is what it already does for new-style declarations).

into a separate PR? I'm not sure why I made this as a part of this PR and not a separate one in the first place, as it was a year ago, but I think it wouldn't be bad to allow some of the new-style rules for the old-style prototypes, especially as this shouldn't break any existing code.

Y-Less:
> 
> 
> > Are there any good examples of a function that legitimately needs to be both, and where this is by far the best solution?
> 
> Don't want to sound rude, but I thought you know where this can be used, as it was you who initially [suggested this](https://github.com/pawn-lang/compiler/issues/306#issuecomment-397771620):
> 

I know, that doesn't mean it was a good idea.  The implementation showed some issues, and on reflection I don't think it is very useful, even if I did suggest it.


> 
> 
> Also, if you're going to close this PR, would you mind if I moved this part
> 
> > * Allows to re-declare functions after their use with old-style prototypes (for some reason this is already possible for new-style prototypes, but not for the old-style ones) (see #438).
> > 

Sounds good

PR #461, opened 2019-10-06
`__nameof`

### LINKED ISSUE #348: nameof operator

* [ ] Bug Report
* [x] Feature Request
* [ ] Question

**What happened**:

Functions like `SetTimer` or `CallLocalFunction` expect their name argument to be a valid public function, but there are currently no mechanisms in Pawn that would prevent common errors associated with this method, i.e. changing a function's name but forgetting to change it in other places where its name is used as a string.

The `nameof` operator (inspired by C#) should expect a symbol as its operand, attempt to find the symbol (similarly to `defined` in processing instructions) and produce the name of the symbol as a string literal. Valid operands should be functions and variables (any kind), but not tags, labels, complex expressions (calls, array access etc.) or numbers.

I think constants (and enums and enum fields) should be also allowed, but macros should be expanded before evaluating the operator. I am not sure how the compiler handles this, but it could be an issue or a subject of further debate.

```pawn
func(arg)
{
    if(arg < 0) return print("`arg` must not be negative");
}
```
Changing `arg` to some other name will require updating its use in other places, but there's a good chance one will forget changing the error message. A partial solution is to use the `#` operator:
```pawn
func(arg)
{
    if(arg < 0) return print("`" #arg "` must not be negative");
}
```
This makes `#arg` harder not to notice, but gives a false sense of security, since the name is still not validated in any way. Assuming the operator could concatenate with literal strings, this would fix the issue:
```pawn
func(arg)
{
    if(arg < 0) return print("`" nameof(arg) "` must not be negative");
}
```

Other common issues are hard-to-notice mistakes in the names:
```pawn
public ChangeColour(playerid)
{
    ...
}

...

SetTimerEx(#ChangeColor, 1000, false, "d", playerid);
```
Debugging the code will sooner or later lead to the issue, but this could be prevented with the new operator very easily:
```pawn
SetTimerEx(nameof ChangeColor, 1000, false, "d", playerid); // error 017: undefined symbol "ChangeColor"
```

This operator would prevent a number of issues associated with calling functions via their name, be friendly with syntax highlighters (and IntelliSense-like tools), and fit well with the other operators like `sizeof` and `tagof` (which also don't evaluate their operands).

**Environment**:

* Operating System
* Compiler version

Y-Less:

`__nameof` See #348

Fixes #348 

* [ ] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:


AGraber:
Can it be done so you can use functions before their definition?

PR #462, opened 2019-10-07
fix "#pragma option" crash when character count is over 31

### LINKED ISSUE #359: #pragma option limited to 31 characters
### Issue description:

Using an option over 31 characters (I assume 32 with NULL) crashes the compiler.

```pawn
// It doesn't even matter if this exists or not.
#pragma option -idependencies\y_thirdpartyinclude
```

### Workspace Information:

* Compiler version:
* Operating System:

AmyrAhmady:

- Title says it all!

Fixes #359

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

For some reason `i` will be `32` after our for loop so an OOB happens when code tries to assign `'\0'` to `name[32]` since `name` size is 32.
Doing a `-1` to `i` fixed this issue and now throws 
```
error 038: extra characters on line
```
when character count is over 31.

Also I'm not sure if I should change `name` size and make it able to accept more than 31 characters, since file paths can be more than 31 characters of course

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1342,14 +1342,9 @@ static int command(void)
           /* first gather all information, start with the tag name */
           while (*lptr<=' ' && *lptr!='\0')
             lptr++;
-          for (i=0; i<sizeof name && *lptr>' '; i++,lptr++)
+          for (i=0; i<sNAMEMAX && *lptr>' '; i++,lptr++)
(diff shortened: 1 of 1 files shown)


Y-Less:
Simpler fix: Instead of `i<sizeof name` use `i<sNAMEMAX`.

PR #463, opened 2019-10-07
Add #pragma once

### LINKED ISSUE #419: #pragma once
### Issue description:

Pretty obvious.  This used to be implicit via `_inc_x`.  Add a pragma that does exactly what the old compiler does:

```pawn
#pragma once
```

Becomes:

```pawn
#if defined _inc_filename
	#endinput
#endif
#define _inc_filename
```

AmyrAhmady:

- Adds `#pragma once`, like old compiler, now you can use `#pragma once` to include the file only once

Fixes #419 

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:
Has compatibility for windows paths and directory separator

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1232,30 +1232,36 @@ static int command(void)
           preproc_expr(&val,NULL);
           sc_needsemicolon=(int)val;
         } else if (strcmp(str,"once")==0) {
-          char* curr_char;
-          char* basefname=inpfname;
-          char symname[sNAMEMAX];
-          symbol *included;
-          /* make it compatible for Windows paths with '\\' directory separator */
-          for (curr_char=inpfname; *curr_char!='\0'; curr_char++)
-            if (*curr_char=='/'||*curr_char=='\\')
-              basefname=curr_char+1;
-          /* assign '_inc_includename' string to symname */
-          strcpy(symname,"_inc_");
-          strlcat(symname,basefname,sizeof symname);
-          /* check if '_inc_includename' is already in global symbols table or not */
-          included=find_symbol(&glbtab,symname,fcurrent,-1,NULL);
-          if (included==NULL)
-            /* couldn't find '_inc_includename' in symbols table */
-            add_constant(symname,1,sGLOBAL,0); // add symname ('_inc_includename') to global symbols table
-          else {
-            /* found '_inc_includename' symbol in global symbols table  */
-            if (!SKIPPING) {
-              assert(inpf!=NULL);
-              if (inpf!=inpf_org)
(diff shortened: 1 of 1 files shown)


AmyrAhmady:
Seems like AppVeyor build failed for some reason, it works well on my machine (Win10, VS 2017)
Also Travis CI build passed, would be glad if someone find me what's wrong.

source/compiler/sc2.c
```diff
@@ -1231,6 +1231,31 @@ static int command(void)
...
+          char symname[sNAMEMAX];
+          symbol *included;
+          /* make it compatible for Windows paths with '\\' directory separator */
+          char* basefname=inpfname;
+          for (char* _cchar=inpfname; *_cchar!='\0'; _cchar++)
+            if (*_cchar=='/'||*_cchar=='\\')
+              basefname=_cchar+1;
+          /* assign '_inc_includename' string to symname */
+          strcpy(symname,"_inc_");
+          strlcat(symname,basefname,sizeof symname);
```
Y-Less: How closely does this replicate all the behaviour here: https://github.com/pawn-lang/compiler/wiki/File-Inclusion

AmyrAhmady:
@Y-Less it only adds `_inc_includefilename.extention` to table, now that I read wiki, seems like I'm doing it wrong.
For example, you can have two `main.inc` in different directories and that way the second one just gets ignored.
What's your idea about this one? It can be helpful.

AmyrAhmady:
### Now what it does is:

with these lines of code:
```pawn
#include <a_samp>
#include <ColAndreas>
#include <ColAndreas>
#include <float>
#include <ColAndreas>
#include <float>
#include "../../ColAndreas"
```
that I added `#pragma once` to `colandreas.inc` and `float.inc`, they only get included once.
and these are my debug messages:
```
Symbol: _inc_float.inc | Added to table
Symbol: _inc_ColAndreas.inc | Added to table
Symbol: _inc_ColAndreas.inc | Already in table
Symbol: _inc_float.inc | Already in table
Symbol: _inc_ColAndreas.inc | Already in table
Symbol: _inc_float.inc | Already in table
Symbol: _inc_../../ColAndreas.inc | Added to table
```

There's also one more thing to mention, `#pragma once` does not work when compiler is running with `-Z` arg; Tell me if I'm doing it wrong.

YashasSamaga:
I think it's better if the include guard symbol does not contain the separators. We need some abstract OS independent symbol name. I think it would be even better to avoid using symbols altogether and use global list of includes.

continue98:
How did this PR end? 

@Y-Less @YashasSamaga

PR #480, opened 2020-02-02
[MERGED] Improvements for detection of unused assignments

### LINKED ISSUE #479: warning 204 only works on unused variables
### Issue description:

Warning 204 (`symbol is assigned a value that is never used`) only works in simple cases when the variable is unused.
```Pawn
main()
{
	new n = 1; // warning 204: symbol is assigned a value that is never used: "n"
}
```
But if an unused value is assigned after the point where the variable is used, then the compiler won't warn about it:
```Pawn
main()
{
	new n = 1;
	printf("%d", n);
	n = 2; // no warning, even though this assignment is unused
}
```
Another problem is when the previously assigned value is unused, the compiler doesn't warn about this kind of errors at all:
```Pawn
new x = 1;
x = 2; // again, no warning, even though the previous value (1) is not used
```
Since the compiler already warns about unused assignments for unused variables, I think it would be beneficial to make it issue warnings in the aforementioned cases as well.

See above.

### Workspace Information:

* Compiler version:
* Operating System:

Daniel-Cortez:

This PR does the following:
* Makes the compiler detect unused assignments that happen after the point where the variable is used:
```Pawn
main()
{
    new n = 1;
    printf("%d", n);
    n = 2; // warning 204: symbol is assigned a value that is never used: "n"
}
```
* Adds warning 240 (`previously assigned value is never used`).
```Pawn
new x = 1;
x = 2; // warning 240: previously assigned value is never used (symbol "x")
```

Fixes #479

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

Currently warning 240 doesn't work on arrays, as it would require a lot more modifications to track assignments to each array element individually.
It is also worth noting that some of the code parts in this PR were copied from the implementation of [warning 210](https://github.com/pawn-lang/compiler/pull/440), as both of them work by the same principle.


Y-Less:
I've not read the code fully.  Does this handle `operator=`?  There should probably be no warning in that case, since you are technically calling a function, which may have desirable side-effects.

Daniel-Cortez:
Not sure what you mean exactly. Can you give an example so I could check it (and see what I can do about it, if necessary)?

Y-Less:
```pawn
Tag:operator=(Tag:a)
{
    printf("hi");
    return Tag:(_:a * 2);
}

main()
{
    new Tag:a = 5, Tag:b;
    // Unused assignment that does something.
    b = a;
}
```

Daniel-Cortez:
Done, such cases are handled now without printing warning 204.

Daniel-Cortez:
* Rebased to resolve merge conflicts.
* Improved detection of unused assignments for function arguments:
```Pawn
DoSomething(arg) // Now 'arg' is marked as `uASSIGNED` ("assigned a value
{                // that is not used yet") upon the beginning of function body.
	arg = 1; // warning 240: previously assigned value is never used
	         // (previously the compiler couldn't detect this).
	#pragma unused arg
}
```

PR #488, opened 2020-02-10
Allow to use `tagof(Tag:)` as a default argument value

### LINKED ISSUE #485: `tagof(Tag:)` doesn't work as a default argument.
### Issue description:

```pawn
MyFunc(a = tagof (Float:))
{
}
```

I tried many things to get it to work.  I couldn't.  Gives:

`
error 001: expected token: "-identifier-", but found "-label-"
`

Daniel-Cortez:

Allows to use `tagof(Tag:)` as a default value for function arguments (see #485).

Fixes #485

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -249,6 +249,7 @@ typedef struct s_symbol {
 #define flagNAKED     0x10  /* function is naked */
 #define flagPREDEF    0x20  /* symbol is pre-defined; successor of uPREDEF */
 
+#define uTAGOF_TAG 0x20  /* set in the "hasdefault" field of the arginfo struct */
 #define uTAGOF    0x40  /* set in the "hasdefault" field of the arginfo struct */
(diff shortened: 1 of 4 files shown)


Daniel-Cortez:
@Y-Less Before reviewing the patch you'll probably want to look at the comments in the test code, they explain why I had to add extra code in `argcompare()` and reserve the 5'th bit in the `arginfo::hasdefault` field. (Basically, the problem is that `tagof(Tag:)` may return different values on the 1'st, the 2'nd and the 3'rd (writing the output to the assembly file) compiler passes.)

Y-Less:
Yes, I saw that comment.  It makes sense, and does explain some earlier bugs that I think are now fixed - namely that of using `tagof` on a tag not actually used anywhere else.

PR #507, opened 2020-04-06
Docker feature

Misiur:
Docker file, all you have to do to get fresh files is to run the container and execute `copy.sh`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,32 @@
+FROM ubuntu:eoan AS compiler-build
+
+RUN apt-get update && \
+    dpkg --add-architecture i386 && \
+    apt-get install -y gcc \
+                       gcc-multilib \
+                       make \
+                       cmake \
+                       git \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+FROM ubuntu:eoan AS compiler
+RUN mkdir /home/compiler
+RUN mkdir /home/build
+WORKDIR /home/build
+COPY --from=compiler-build /home/build/pawncc .
+COPY --from=compiler-build /home/build/pawnruns .
+COPY --from=compiler-build /home/build/pawndisasm .
+COPY --from=compiler-build /home/build/libpawnc.so .
+WORKDIR /home
+ADD ./copy.sh /home
```
Southclaws, line 30: What's the point of this script?

Misiur: Mounted `/compiler` directory via compose to copy artifacts

Issue #464, opened 2019-10-07
Use `operator=` on tag-mismatched returns.


Y-Less:
### Minimal complete verifiable example (MCVE):

```pawn
Tag:operator=(a) return Tag:a;

Tag:Func()
{
    return 7;
}
```

That is currently a tag-mismatch, even though it is pretty much an assignment.

### Workspace Information:

* Compiler version:
* Command line arguments provided (or sampctl version):
* Operating System:

Y-Less:
For reference, this code calls the operator even though there's no explicit `=` here either:

```
Tag:operator=(a) return Tag:a;

Func(Tag:a)
{
}

main()
{
    Func(7);
}
```

I don't think the two situations are fundamentally different.

PR #519, opened 2020-05-03
Fix user-defined assignment operators not being called for compound assignments

### LINKED ISSUE #517: Missing operator call after compound assignment operator (+=, -=, ...)
### Issue description:

The assigment operator isn't called for a tag mismatch after a compound assignment operator
If the tag matches or if a constant value, maybe other further cases I didn't check

If I am not mistaken this also didn't work in the old compiler so it is a pretty old bug I guess maybe it was fixed in any newer pawn version

```
stock Tag: operator=(oper) {
    return Tag: oper;
}

stock operator+(Tag: oper1, Tag: oper2) {
    return (_: oper1 + _: oper2);
}

stock operator+(Tag: oper1, oper2) {
    return (_: oper1 - _: oper2);
}

main() {
    new Tag: t1, Tag: t2;
    // calls Tag: operator=(oper) after addition
    t1 = t1 + t2; // operator+(Tag:, Tag:)
    t1 = t1 + 5; // operator+(Tag:, _:)
    t1 += _: t2; // operator+(Tag:, _:)
    // does not call Tag: operator=(oper) after addition
    t1 += t2; // operator+(Tag:, Tag:)
    t1 += 5; // operator+(Tag:, _:)
}
```

### Workspace Information:

* Compiler version: 3.10.10
* Command line arguments provided (or sampctl version): sampctl version 1.8.39
* Operating System: Windows

Daniel-Cortez:

Fixes the compiler not calling user-defined assignment operator for compound assignments, as described in #517.

Fixes #517 

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -1053,7 +1053,8 @@ static int hier14(value *lval1)
   if (leftarray) {
     memcopy(val*sizeof(cell));
   } else {
-    check_userop(NULL,lval2.tag,lval3.tag,2,&lval3,&lval2.tag);
+    if (check_userop(NULL,(oper==NULL) ? lval2.tag : lval1->tag,lval3.tag,2,&lval3,&lval2.tag))
+      lval1->tag=lval2.tag; /* user-defined assignment operator can override the resulting tag */
     store(&lval3);      /* now, store the expression result */
   } /* if */
   if (!oper)
--- a/source/compiler/tests/gh_517.meta
+++ b/source/compiler/tests/gh_517.meta
@@ -1,0 +1,5 @@
+{
+  'test_type': 'output_check',
+  'errors': """
+"""
+}
--- a/source/compiler/tests/gh_517.pwn
+++ b/source/compiler/tests/gh_517.pwn
@@ -1,0 +1,20 @@
+stock Tag:operator=(oper)
+	return Tag:oper;
+
+stock operator+(Tag:oper1, Tag:oper2)
+	return (_:oper1 + _:oper2);
+
+stock operator+(Tag:oper1, oper2)
+	return (_:oper1 - oper2);
+
+main()
+{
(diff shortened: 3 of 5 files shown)


Daniel-Cortez:
A bit of explanation about why I replaced `lval2.tag` with `(oper==NULL) ? lval2.tag : lval1->tag`: as it turns out, the tag for the assigned value is stored in `lval2.tag` only for simple assignments (`=`), and for compound assignments (`+=`, `-=`, ...) the tag is actually in `lval1->tag`, because the addition/subtraction/whatever operation, which is handled before the assignment, stores the resulting tag there first. This is why the compiler didn't call `Tag:operator=(oper)` in #517 in some of the compound assignments, because when trying to find the user-defined assignment operator it passed a tag from the wrong value to `check_userop()`.

Also, user-defined assignment operators can override the resulting tag (e.g. in `Tag:operator=(oper)` the assigned value has no tag, but the result has tag `Tag:`), so I added an extra line of code to copy the tag into `lval1->tag` if user-defined assignment takes place.
This fixes another bug described in https://github.com/pawn-lang/compiler/issues/517#issuecomment-623142645

I also added two tests: the first one is a P-code check to make sure the the operators are called correctly, and the other one simply checks if there are no new tag mismatch warnings.

Daniel-Cortez:
I fixed the tests, but for some reason Travis CI doesn't seem to respond. Except for that, the PR should be ready now.

Southclaws:
It definitely passed, but it seems the GitHub integration is stuck.

https://travis-ci.org/github/pawn-lang/compiler/builds/682961252

Y-Less:
So I was going to bring up this issue as well, if you're looking at places where assigmnent operators should be called:

https://github.com/pawn-lang/compiler/issues/464

PR #520, opened 2020-05-05
Use `operator=()` on tag-mismatched returns

### LINKED ISSUE #464: Use `operator=` on tag-mismatched returns.

Daniel-Cortez:

See #464 

Fixes #464

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -7588,8 +7588,6 @@ static void doreturn(void)
         error(90,curfunc->name);        /* public function may not return array */
     } /* if */
     rettype|=uRETVALUE;                 /* function returns a value */
-    /* check tagname with function tagname */
-    check_tagmismatch(curfunc->tag,tag,TRUE,-1);
     if (ident==iARRAY || ident==iREFARRAY) {
       int dim[sDIMEN_MAX],numdim=0;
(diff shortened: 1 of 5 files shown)


Daniel-Cortez:
As with #519, I added two tests: one checks that there are no tag mismatch warnings (and also that one of the three operators is unused, as it's not supposed to be used on tag-matching returns), and the other one is a P-code test that makes sure the operators are called properly.

Another thing worth mentioning is that in the P-code test I couldn't check which operator is called in which case exactly. The names of non-public functions, including user-defined operators, are only present in assembly listings from the compiler (obtained with the `-a` option), but not in disassembler output, and function addresses are not reliable enough to be used in tests.
On the other hand, both operators (`Tag:operator=(a)` and `operator=(Tag:a)`) are defined without the `stock` specifier and the compiler doesn't warn about them being unused, so at least that (sort of) confirms that they're used.

PR #521, opened 2020-05-05
Fix anonymous enum elements enforcing their tag

### LINKED ISSUE #514: tag mismatch
### Issue description:
Assigning a tagged value to an array cell indexed with an enum constant makes the compiler print warning 213 (tag mismatch), even though the array has the same tag.

```Pawn
#include <a_samp>

const TD_0 = 0;
enum { TD_1 = 1 };
new Text:textdraws[2];

main()
{
    // no warnings/errors on the following line, as `TD_0`
    // is a regular constant, not an enum element
    textdraws[TD_0] = TextDrawCreate(0.0, 0.0, "-");

    // warning 213: tag mismatch: expected tag none ("_"), but found "Text"
    textdraws[TD_1] = TextDrawCreate(0.0, 0.0, "-");
}
```
It seems that if an array is indexed with an enum constant, the tag of the said constant somehow overrides the array tag, so in the above example the compiler expects no tag (`_:`) instead of `Text:`, which doesn't seem to be correct.

Originally reported by the forum user 'tnc' here: https://pro-pawn.ru/showthread.php?2207-Pawn-compiler-%283-10%29&p=96537&viewfull=1#post96537

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.1 - 3.10.10
* Operating System:

Daniel-Cortez:

Fixes anonymous enum elements overriding the resulting tag when they're used for array indexing, as described in #514.

Fixes #514

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -1776,11 +1776,11 @@ restart:
           } /* if */
           charalign();                  /* align character index into array */
         } /* if */
-        /* if the array index is a field from an enumeration, get the tag name
-         * from the field and save the size of the field too.
+        /* if the array index is a field from a named enumeration, get the tag
+         * name from the field and save the size of the field too.
          */
         assert(lval2.sym==NULL || lval2.sym->dim.array.level==0);
-        if (lval2.sym!=NULL && lval2.sym->dim.array.length>0 && sym->dim.array.level==0) {
+        if (lval2.sym!=NULL && lval2.sym->parent!=NULL && lval2.sym->dim.array.length>0 && sym->dim.array.level==0) {
           lval1->tag=lval2.sym->x.tags.index;
           lval1->constval=lval2.sym->dim.array.length;
         } /* if */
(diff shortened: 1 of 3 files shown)


source/compiler/sc3.c
```diff
@@ -1832,8 +1832,6 @@ static int hier1(value *lval1)
       if (lval2.ident==iCONSTEXPR && lval2.sym!=NULL
           && lval2.sym->dim.array.length>0 && sym->dim.array.level==0)
       {
-        lval1->tag=lval2.sym->x.tags.index;
```
Daniel-Cortez, line 1835: I removed these two lines, as they turned out to be duplicate code. There's already code that does the same thing (saves the tag and the enum field size into the resulting value) at lines 1784-1785.

source/compiler/sc3.c
```diff
@@ -1832,8 +1832,6 @@ static int hier1(value *lval1)
...
-        lval1->constval=lval2.sym->dim.array.length;
         if (lval2.tag==sym->x.tags.index && lval1->constval>1 && matchtoken('[')) {
```
Daniel-Cortez, line 1835: I removed these two lines, as they turned out to be duplicate code. There's already code that does the same thing (saves the tag and the enum field size into the resulting value) at lines 1784-1785.

source/compiler/sc3.c
```diff
@@ -1832,8 +1832,6 @@ static int hier1(value *lval1)
       if (lval2.ident==iCONSTEXPR && lval2.sym!=NULL
           && lval2.sym->dim.array.length>0 && sym->dim.array.level==0)
       {
-        lval1->tag=lval2.sym->x.tags.index;
-        lval1->constval=lval2.sym->dim.array.length;
```
Daniel-Cortez, line 1836: I removed these two lines, as they turned out to be duplicate code. There's already code that does the same thing (saves the tag and the enum field size into the resulting value) at lines 1784-1785.

PR #526, opened 2020-05-16
Operator `__pragma`

### LINKED ISSUE #267: __attribute__((attr-list)) or [[attr-list]]

Daniel-Cortez:

This PR implements operator `__pragma` (see #267).

Currently available options:
* `deprecated`
* `unused`
* `unread`
* `unwritten`
* `nodestruct`
* `naked`

Examples of use:
```Pawn
native Func() __pragma("deprecated - use OtherFunc() instead");

stock __pragma("naked") NakedFunc2() {}

NakedFunc()
{
    __pragma("naked", "deprecated - use NakedFunc2() instead");
}

new __pragma("nodestruct") Tag:b = Tag:0;
new a = 0 __pragma("unused");
new __pragma("unread") c = 0 __pragma("unwritten");
new d = 0 __pragma("unread", "unwritten");
```

Fixes #267

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -400,6 +400,7 @@ enum {
   tNATIVE,
   tNEW,
   tOPERATOR,
+  t__PRAGMA,
   tPUBLIC,
   tRETURN,
   tSIZEOF,
@@ -571,6 +572,16 @@ enum {  /* search types for error_suggest() when the identifier type is "estSYMB
   esfVARCONST   = esfCONST | esfVARIABLE | esfARRAY
 };
 
+enum { /* attribute flags for "__pragma" */
+  attrDEPRECATED,
+  attrUNUSED,
+  attrUNREAD,
+  attrUNWRITTEN,
+  attrNODESTRUCT,
+  attrNAKED,
+  NUM_ATTRS
+};
+
 /* interface functions */
 #if defined __cplusplus
   extern "C" {
@@ -659,6 +670,9 @@ SC_FUNC symbol *add_builtin_string_constant(char *name,const char *val,int vclas
 SC_FUNC void exporttag(int tag);
 SC_FUNC void sc_attachdocumentation(symbol *sym);
 SC_FUNC void emit_parse_line(void);
+SC_FUNC void pragma_deprecated(symbol *sym);
+SC_FUNC void pragma_unused(symbol *sym,int unread,int unwritten);
+SC_FUNC void pragma_nodestruct(symbol *sym);
 
 /* function prototypes in SC2.C */
 #define PUSHSTK_P(v)  { stkitem s_; s_.pv=(v); pushstk(s_); }
@@ -955,6 +969,7 @@ SC_VDECL int pc_memflags;     /* special flags for the stack/heap usage */
 SC_VDECL int pc_naked;        /* if true mark following function as naked */
 SC_VDECL int pc_compat;       /* running in compatibility mode? */
 SC_VDECL int pc_recursion;    /* enable detailed recursion report? */
+SC_VDECL unsigned int pc_attributes;/* currently set attribute flags (for the "__pragma" operator) */
 
 SC_VDECL constvalue_root sc_automaton_tab; /* automaton table */
 SC_VDECL constvalue_root sc_state_tab;     /* state table */
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -8172,7 +8172,7 @@ static void pragma_apply(symbol *sym)
       break;
     case attrNAKED:
       if (sym->ident==iFUNCTN)
-        sym->flags=flagNAKED;
+        sym->flags |= flagNAKED;
       break;
     default:
       assert(0);
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -5510,6 +5510,11 @@ static void statement(int *lastindent,int allow_decl)
     matchtoken(tSTATIC);
     decl_enum(sLOCAL,FALSE);
     break;
+  case t__PRAGMA:
+    dopragma();
+    needtoken(tTERM);
+    pragma_apply(curfunc);
+    break;
   case t__EMIT: {
     extern char *sc_tokens[];
     const unsigned char *bck_lptr=lptr-strlen(sc_tokens[tok-tFIRST]);
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -782,6 +782,8 @@ cleanup:
                                            * done (i.e. on a fatal error) */
   delete_symbols(&glbtab,0,TRUE,TRUE);
   line_sym=NULL;
+  free(pc_deprecate);
+  pc_deprecate=NULL;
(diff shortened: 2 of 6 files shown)


Y-Less:
> `__pragma("unread", "unwritten");`

That's a good solution to multiple properties.

Y-Less:
Why didn't you reuse the existing `#pragma` code?  It seems it would have been simpler to just use that and add an extra parameter to disable some like `codepage` for `__` compared to `#`.

Daniel-Cortez:
Because the code for `__` is too different from `#`.

* The option names are handled differently.
Since operator `__pragma` can be specified before the symbol is known, e.g.
```
new __pragma("nodestruct") Tag:var;
```
  ... , I had to make the compiler memoize the specified attributes, so they could be applied to the symbol later - the most obvious solution was to store bit flags into a global variable.
This is why while `#pragma` distinguishes the options by their names, `__pragma` uses bit flags when applying the attributes.
Of course, I could just convert the bit flags back into attribute names, but there are still other problems listed below.

* The option names and parameters are read differently.
In `#pragma` the option name and parameters are read directly (by manipulating with `lptr`), but in `__pragma` they are read from the literal queue, because the operator reuses the standard string parsing mechanism.

* The parameters are handled differently.
In `#pragma unused/unread/unwritten/nodestruct` the symbol name is specified after the option name (e.g. `#pragma nodestruct var`), but in `__pragma` the symbol is separate from the operator.

Thus the only thing left to reuse is the code responsible for applying the attributes to symbols (which isn't really that much of code), which could be moved into separate reusable functions:
```C
SC_FUNC void pragma_unused(symbol *sym, int unread, int unwritten)
{
  sym->usage |= unwritten ? 0 : uREAD;
  if (sym->ident == iVARIABLE || sym->ident == iREFERENCE
      || sym->ident == iARRAY || sym->ident == iREFARRAY)
    sym->usage |= unread ? 0 : uWRITTEN;
}

SC_FUNC void pragma_nodestruct(symbol *sym)
{
  if (sym->ident==iVARIABLE || sym->ident==iARRAY)
    sym->usage |= uNODESTRUCT;
}
```

Y-Less:
That makes sense.

Daniel-Cortez:
Done, moved the code for `deprecated`, `unused`, `unread`. `unwritten` and `nodestruct` into reusable functions, as described in the post above.

While I was at it, I also fixed a potential memory leak in `#pragma deprecated` (the memory block for the deprecaton message is not freed if it's unused, which might happen either because of end of script or if there are two deprecation `#pragma`s used in a row).

Y-Less:
I've merged this, but it would be nice if it were possible to use `"naked"` inside functions as well as in their header:

```pawn
Func()
{
    __pragma("naked");
}
```

Daniel-Cortez:
> it would be nice if it were possible to use `"naked"` inside functions as well as in their header

Done.
As a bonus, aside from `"naked"`, the other attributes can be used inside functions as well:
```Pawn
NakedFunc()
{
	__pragma("naked", "deprecated - use NakedFunc2() instead");
}
```

Y-Less:
Oh, I just discovered that `naked` removes `PROC` as well.  That's good to know.

Y-Less:
The tests don't really cover that, but they're only testing compiler warnings.  Is that actually correct?

Daniel-Cortez:
> The tests don't really cover that, but they're only testing compiler warnings. Is that actually correct?

If by "that" you mean all other attributes not being tested inside functions, then I think it might unnecessarily bloat the tests. As you can see it in ab66969269f6f7e3834706f8608a8e42235fbdea, the only difference is in the code that calls functions `dopragma()` and `pragma_apply(curfunc)`, making it possible for `__pragma` to be used inside functions; the code that reads and applies the attributes is still the same, which is why I added only one basic check to make sure the syntax is being handled correctly. But I can expand the tests if you think that check isn't enough.

Y-Less:
Let me rephrase:

I just discovered that `naked` removes `PROC` as well, is that correct?  The tests only cover compiler warnings, not assembly output, so they don't cover the lack of `PROC`.

I was just talking about `PROC` in both posts.

Daniel-Cortez:
Ah, so that's what you meant.
Alright then, I'll add a P-code test.

EDIT: OK, done: #571.

PR #533, opened 2020-06-13
[MERGED] Further improvements for detection of unused assignments

### LINKED ISSUE #479: warning 204 only works on unused variables

### LINKED ISSUE #545: The warning 204 when assigning in "do{} while();" condition scope
### Issue description:

The following code results in a warning 204, although it should not:

```pawn
stock func() 
    return 1;

main()
{
    new i = 0;
    do
    {
         printf("%d", i);
    } while((i = func()));//symbol is assigned a value that is never used: "i"
}
```

See above.

### Workspace Information:

* Compiler version: 3.10.10
* Operating System:

Daniel-Cortez:

This PR continues the work regarding detection of unused assignments (warnings 204 and 240), the first part of which was done in #480.
* Fixes a bug with assignments inside loops being mistakenly reported as unused after using `break` or `continue`.
```Pawn
new var = 0;
for (/* ... */) {
    if (/* ... */) {
        var = 0;
        break;
    }

    // Before the fix, the compiler falsely reported the previous assignment "var = 0"
    // as unused ("warning 240: previously assigned value is never used (symbol "var")").
    var = var + 1;
}
```

* Improves the unused assignment detection algorithm by tracking the "compound statement" nesting level for each assignment, which allows to detect more complicated cases of unused assignments.

Fixes #479

* [x] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

This PR relies on the fix from #532, so you'll probably want to review and merge that PR first.

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -957,6 +957,7 @@ SC_VDECL int pc_memflags;     /* special flags for the stack/heap usage */
 SC_VDECL int pc_naked;        /* if true mark following function as naked */
 SC_VDECL int pc_compat;       /* running in compatibility mode? */
 SC_VDECL int pc_recursion;    /* enable detailed recursion report? */
+SC_VDECL int pc_nestlevel;    /* number of active (open) compound statements */
 
 SC_VDECL constvalue_root sc_automaton_tab; /* automaton table */
 SC_VDECL constvalue_root sc_state_tab;     /* state table */
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -701,7 +701,6 @@ SC_FUNC int refer_symbol(symbol *entry,symbol *bywhom);
 SC_FUNC void markusage(symbol *sym,int usage);
 SC_FUNC void markinitialized(symbol *sym,int assignment);
 SC_FUNC void clearassignments(symbol *root,int fromlevel);
-SC_FUNC void demoteassignments(symbol* root,int level);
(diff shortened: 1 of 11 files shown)


Daniel-Cortez:
Fixed a bug with false positive warnings 204 and 240 printed with random (garbage) line numbers, e.g.:
```
pawno\include\functions.inc(758132272) : warning 204: symbol is assigned a value that is never used: "playerid"
```
The reason of this problem was [the bug with local constants](https://github.com/pawn-lang/compiler/issues/541) which broke the code in `memoizeassignments()` and `restoreassignments()`.
In order to fix this, I incorporated a fix from #543 into this PR.

Issue #528, opened 2020-05-24
New warnings/errors?


Daniel-Cortez:
### Issue description:
Pawn is positioned as ["a robust language with a compiler that performs a maximum of static checks"](https://www.compuphase.com/pawn/pawn.htm#INTRODUCTION), but there are still a lot of areas where these checks could be improved.
I think it could be useful to have the following warnings/errors:
* **possible use of symbol before initialization**
  This one was already planned by the original author of Pawn as "warning 210" at some point of development (and even documented in the Language Manual), but for some reason the logic for this warning was never implemented. Experimental implementation is available in #440.

* **redundant code: switch argument is a constant value**
  We already have similar warnings for `if` statements:
```Pawn
if (false) {} // warning 205: redundant code: constant expression is zero
if (true) {} // warning 206: redundant code: constant expression is non-zero
```
  so why not adopt the same for `switch`?

* **negative or too big shift count**
  This can be achieved if the user has too many bit flags.

* **not all enum items are covered by switch statement**
  Actually, I already experimented with implementing this warning locally last year, and had a few false-positives in [YSI](https://github.com/pawn-lang/YSI-Includes/blob/5.x/YSI_Coding/y_hooks/y_hooks_impl.inc#L467) and amx_assembly ([1](https://github.com/Zeex/amx_assembly/blob/master/codescan.inc#L426), [2](https://github.com/Zeex/amx_assembly/blob/master/opcode.inc#L270)). I think restricting this warning to cases where only 1 or 2 enum items are missing should help to get rid from this kind of false-positives.

* **division by zero**
  Interestingly, the compiler already has a check for zero division in function `flooreddiv()` (sc3.c), but it only works if both operands are constant values:
```Pawn
new x = 1 / 0; // error 029: invalid expression, assumed zero
new y = x / 0; // no errors/warnings if the dividend is non-constant
```
  Also, the message "`invalid expression, assumed zero`" in the above example is misleading, because
  1) this is not the same as "division by zero";
  2) usually the compiler prints this message because of invalid syntax, while zero division is rather a semantic error.

### Workspace Information:

* Compiler version:
* Command line arguments provided (or sampctl version):
* Operating System:

Daniel-Cortez:
Should I implement those warnings exactly as described? Would be nice to have any feedback before I get to implementing them.

Daniel-Cortez:
Update: Added an idea for a new warning:
> * **enum increment "%s" has no effect on zero value**
[...]

PR #539, opened 2020-06-21
[MERGED] warning 24X: negative or too big shift count

Daniel-Cortez:

This PR allows the compiler to warn the user if the shift count is too big or negative (both of which are UB), as suggested in #528.
For example, this could be useful when the user tries to store too many bit flags into one variable:
```Pawn
enum ePlayerFlags
{
/*  0 */    pfEmailConfirmed,
/*  1 */    pfTutorialCompleted,
            /* ... */
/* 31 */    pfImprisoned,
/* 32 */    pfHospitalized
};

// ...

player_info[playerid][pFlags] |= (1 << pfHospitalized); // warning 24X: negative or too big shift count
```

Fixes #

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:


Y-Less:
What about right shifts?

Also, a related suggestion would be a warning for this:

```pawn

enum A (<<= 1)
{
	B = 1,
	c00,
	c01,
	c02,
	c03,
	c04,
	c05,
	c06,
	c07,
	c08,
	c09,
	c10,
	c11,
	c12,
	c13,
	c14,
	c15,
	c16,
	c17,
	c18,
	c19,
	c20,
	c21,
	c22,
	c23,
	c24,
	c25,
	c26,
	c27,
	c28,
	c29,
	c30,
	c31,
	c32,
	c33,
	c34,
	c35,
	c36,
	c37,
	c38,
	c39
}
```

Where the enum shifts overflow.

Daniel-Cortez:
>Also, a related suggestion would be a warning for this:
>(...)
>Where the enum shifts overflow.

I wanted to do this initially, but currently the left shift in enum declarations is processed in the same way as multiplication (e.g. `<<= 1` is treated as `*= 2` internally, `<<= 2` becomes `*= 4`, etc.), so currently there's no way to catch invalid shift values. I think I'll need to rewrite some of the code that handles increments in function `decl_enum()` (file `sc1.c`) to make the compiler process the left shift operation in a distinct way (which might also open a way to solution for #540).

> What about right shifts?

But it does work on right shifts.

Y-Less:
> But it does work on right shifts.
> **EDIT:** Ah, right, didn't cover them with tests; I'll fix this.

Nice.

Daniel-Cortez:
Updated this PR; now warning 24X is also printed when the specified left shift count is too big
```Pawn
enum (<<= 32) // warning 24X: negative or too big shift count
{
    // ...
};
```

Daniel-Cortez:
Also, even though I incorporated the fix from #546, I haven't made warning 24X work for shift overflow in enum items yet, as I'm not sure if this warning is correct for such situations.
```Pawn
enum (<<= 1)
{
    i0 = 1,
    i1,
    i2,
    /* ... */
    i31,
    i32, // overflow
};
```
In this example, technically, `i32` is not `1 << 32`, it's `i31 << 1`, so the message "negative or too big shift count" doesn't sound correct in this case.
Maybe it would make sense to add a separate warning for overflow in enum items (and probably not only for left shift, but also for multiplication and addition)? And if so, what would be the best way to phrase the warning message? I'm not a native English speaker, so I'm not sure; for now I'm going with "shift overflow for enum item (symbol \"%s\")", but there's probably a better or more correct way to phrase it.

PR #567, opened 2020-09-03
[MERGED] Fix the array return bug

### LINKED ISSUE #293: Nested array returns cause bad memory/invalid instructions

* [x] Bug Report
* [ ] Feature Request
* [ ] Question

**What happened**:

Returning an array into another return into the arguments of another function result in a runtime crash.

No runtime crash, I expect to nest functions in arguments any number of times with no issues.

With: https://github.com/sampctl/pawn-array-return-bug

```powershell
sampctl get sampctl/pawn-array-return-bug
cd pawn-array-return-bug
sampctl package run --forceBuild
```

I'm not entirely sure if this is something that can be fixed via the compiler but from my limited knowledge, I think it could be fixed.

People run into this a lot when they cut corners by writing a `GetPlayerName` wrapper that returns the name instead of writing it into a pass-by-reference. I often advise people to avoid doing this, but it still seems like something that should be fixed if it can be.

**Environment**:

* Operating System: All
* Compiler version: 3.10.4

Daniel-Cortez:

Fixes bad code generation when returning an array returned by another function (see #293).

Fixes #293

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

The test code is based on the [array return bug demonstration](https://github.com/sampctl/pawn-array-return-bug) by @Southclaws.

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -955,6 +955,8 @@ SC_VDECL int pc_memflags;     /* special flags for the stack/heap usage */
 SC_VDECL int pc_naked;        /* if true mark following function as naked */
 SC_VDECL int pc_compat;       /* running in compatibility mode? */
 SC_VDECL int pc_recursion;    /* enable detailed recursion report? */
+SC_VDECL int pc_retexpr;      /* true if the current expression is a part of a "return" statement */
+SC_VDECL int pc_retheap;      /* heap space (in bytes) to be manually freed when returning an array returned by another function */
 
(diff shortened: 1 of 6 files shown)


afwn90cj93201nixr2e1re:
Is that produce memory leaks in run-time?

Y-Less:
> Is that produce memory leaks in run-time?

It doesn't leak, no.  Pawn itself doesn't really have memory leaks, those all come from plugins.

PR #572, opened 2020-09-21
Fix false-positive warning 242 messages

Daniel-Cortez:

* Fixes false-positive warning 242 messages when the enum increment operator is `<<=` and the shift count is `0`:
```Pawn
enum (<<= 0)
{
    eItem1 = 1,
    eItem2 // warning 242: shift overflow in enum item declaration (symbol "eItem2")
};
```
* Disables warning 242 for situations when the shift count for the enum increment is too big:
```Pawn
enum (<<= 32) // warning 241: negative or too big shift count
{
    eItem1 = 1,
    eItem2, // warning 242: shift overflow in enum item declaration (symbol "eItem2")
    eItem3  // warning 242: shift overflow in enum item declaration (symbol "eItem3")
    // While the warning 242 messages are correct in this case, we don't really need them,
    // as the compiler already warned us about invalid shift count earlier with warning 241.
};
```

Fixes #

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:


Y-Less:
One thing I just noticed - this (and warning 244) should say "enum field", not "enum item".  That's the terminology used in the documentation.

And maybe change 241 to "invalid enum modifier" (or "invalid enum increment"), so it can cover a wider range of cases, such as `*= 0`, `>>>= 0` etc (I believe those are invalid, but maybe someone has a use-case for them).  Maybe also a warning for when enum fields stop updating, like:

```pawn
enum E (>>= 1)
{
    A = 4,
    B, // 2
    C, // 1
    D, // 0
    E, // Also 0, warning here.
}
```

Daniel-Cortez:
> One thing I just noticed - this (and warning 244) should say "enum field", not "enum item". That's the terminology used in the documentation.

IIRC, I picked the term "enum item" from this comment:
https://github.com/pawn-lang/compiler/blob/134ad7a836c581546665340aedb59efd4636e269/source/compiler/sc.h#L152
and I don't think replacing this term with "enum field" would be a good idea.
From how I understand it, the term "enum field" implies `enum` being used as `struct` from C/C++, while "enum item" has a more general meaning.
Also, the only two mentions of "enum field" in the documentation are in the "Error and warning messages" section
> 227 more initiallers than **enum fields**
> An array whose size is declared with an enum symbol contains (...)

> 228 length of initialler exceeds size of the **enum field**
> An array whose size is declared with an enum symbol, and (...)

which are indeed for the cases where enumerations are used as array sizes.
I think all mentions of "enum item" should be left as they are.

> And maybe change 241 to "invalid enum modifier" (or "invalid enum increment"), so it can cover a wider range of cases, such as `*= 0`, `>>>= 0` etc (I believe those are invalid, but maybe someone has a use-case for them)

In Pawn only `+=`, `*=` and `<<=` are allowed for enum increments, so `*= 0` would be valid.
Regarding warning 241, it's also used outside of enums:
```Pawn
new y = x >> 32; // warning 241: negative or too big shift count
```
so changing its phrasing to "invalid enum increment" wouldn't be valid for those cases.
I guess I could create a separate warning for too big multipliers used as enum increments, but I don't know how to phrase such warning in the most descriptive way. Any ideas?

And I'm sure I could create another warning to cover multiplication overflows for enum items:
```Pawn
enum (*= 1024)
{
    eField1 = 1,
    eField2, // 1024
    eField3, // 1048576
    eField4, // 1073741824
    eField5  // warning 245: multiplication overflow in enum item declaration (symbol "eField5")
};
```

Y-Less:
There are far more mentions:

> array initialization and enumerations
> The array size may be set with a constant that represents an enumeration:
> an example of this is the “priority queue” sample program at page 21. When
> individual fields of the enumeration have a size, the associated array elements
> are interpreted as sub-arrays, on occasion. For an example of this behaviour,
> see the rpn calculator program at page 30.

> The initialler consists of a string (a literal array) and a value; these go into the
> fields “text” and “priority” respectively.

> A typical use of “tagged” enum’s is in conjunction with arrays. If every field of
> an array has a distinct purpose, you can use a tagged enum to declare the size
> of an array and to add tag checking to the array usage in a single step:

Y-Less:
Or some other name, but "item" seems really wrong.  Very non-descriptive.

Daniel-Cortez:
> There are far more mentions:
> (...)

I just checked all mentions of the word "field" alone in pawn-lang.pdf, and, again, all of them seem to be about enumerations being used as array sizes (with a few exceptions where this word is used in a context not related to enumerations).

> Or some other name, but "item" seems really wrong. Very non-descriptive.

Well, I'm not against replacing the term "enum item" in general, it's just that "enum field" seems to have a too specific meaning to be a fitting replacement.

The documentation also uses the terms "enum constant" (which, again, sounds too specific) and "enum element"; maybe the latter could be a good candidate?

Daniel-Cortez:
Also, speaking of incorrect terms, I recently noticed that in the compiler code the word "semicolumn" is oftenly used as a synonym of "semicolon" - is that even correct? If not, then we should probably fix it.

**EDIT:** Just tried to google for "semicolumn"; the engine automatically corrected it to "semicolon", which suggests that it might indeed be a common mistake.

Y-Less:
No, that's not a word.  Probably just a typo.

Y-Less (approved):

PR #578, opened 2020-09-30
Compare the contents of default arrays between function declaration(s) and implementation

### LINKED ISSUE #576: No warnings when default array values differ between function declaration(s) and implementation
### Issue description:
The compiler warns when default values of function arguments differ between declaration(s) and implementation:
```Pawn
forward Func(x = 0);

stock Func(x = 1) // error 025: function heading differs from prototype
{
	return x;
}
```
But for some reason it doesn't warn when default values for array arguments are different:
```Pawn
forward Func(const a[] = { 1, 2, 3 });

stock Func(const a[] = { 3, 4, 5 }) // no warning(s)
{
	return a[0];
}
```

See above.

### Workspace Information:
* Operating System:

Daniel-Cortez:

Fixes the compiler not warning when default values of array arguments differ between declaration(s) and implementation (see #576).

Fixes #576

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -4048,9 +4048,8 @@ static int argcompare(arginfo *a1,arginfo *a2)
         result= a1->defvalue.array.size==a2->defvalue.array.size;
       if (result)
         result= a1->defvalue.array.arraysize==a2->defvalue.array.arraysize;
-      /* ??? should also check contents of the default array (these troubles
-       * go away in a 2-pass compiler that forbids double declarations, but
-       * Pawn currently does not forbid them) */
+      if (result)
+        result=(memcmp(a1->defvalue.array.data,a2->defvalue.array.data,a1->defvalue.array.size*sizeof(cell))==0);
     } else {
       if (result) {
         if ((a1->hasdefault & uSIZEOF)!=0 || (a1->hasdefault & uTAGOF)!=0)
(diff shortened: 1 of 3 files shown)


source/compiler/sc1.c
```diff
@@ -4048,9 +4048,8 @@ static int argcompare(arginfo *a1,arginfo *a2)
         result= a1->defvalue.array.size==a2->defvalue.array.size;
       if (result)
         result= a1->defvalue.array.arraysize==a2->defvalue.array.arraysize;
-      /* ??? should also check contents of the default array (these troubles
-       * go away in a 2-pass compiler that forbids double declarations, but
-       * Pawn currently does not forbid them) */
```
YashasSamaga, line 4053: I am unable to understand this comment. What were the original authors trying to say here?

Daniel-Cortez: By "double declarations" the original author probably meant this:
```Pawn
forward Func(); // declared
forward Func(); // declared again, not an error
forward Func(arg); // error 025: function heading differs from prototype
```
Pawn allows functions to be forward-declared more than once, as long as all declarations are identical.

Y-Less: That's actually quite essential to some code.  There's a technique that actually abuses this and the definition of "identical" just being semantic.  These two are the same:

```pawn
Func(x[_:A:B:C:0]);
Func(x[_:D:E:F:0]);
```

So that code will compile with no issues, and is thus, as with all weird code that compiles, abused...

Issue #566, opened 2020-08-29
Code after `exit` statements is not detected as unreachable


Daniel-Cortez:
### Issue description:

Operator `exit` aborts execution, so I believe it would make sense to allow the compiler to detect unreachable code after `exit` statements.

### Minimal complete verifiable example (MCVE):

```Pawn
main()
{
    exit; // works as a shortcut for "exit 0;"
    return 0; // no warning(s), although this code is unreachable
}
```

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Daniel-Cortez:
Fixed in #580.

Issue #565, opened 2020-08-29
Labels are falsely detected as unreachable code


Daniel-Cortez:
### Issue description:

If a label is used via `goto` and then defined after an `if` statement with both branches ending with `return`, the compiler thinks the label is unreachable.

### Minimal complete verifiable example (MCVE):

```Pawn
main()
{
    new var = 0;
    if (var == 0)
        goto label_1;
    if (var != 0)
        return 0;
    else
        return 1;
label_1: // warning 225: unreachable code
    return 2;
}
```

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Daniel-Cortez:
Fixed in #580.

Issue #560, opened 2020-08-07
Unreachable code after `switch` is not being detected


Daniel-Cortez:
### Issue description:

Currently the compiler isn't able to detect unreachable code after `switch` statements when all cases (including `default`) end with `return`.
```Pawn
Func(x) return x;
main()
{
    new x = 0;
    switch (x)
    {
        case 0: return Func(0);
        default: return Func(1);
    }
    return Func(2); // no warning(s), even though
                    // the code on this line is unreachable
}
```
Since the compiler already can detect unreachable code when both `if` branches end with `return`
```Pawn
main()
{
    new x = 0;
    if (x)
        return 0;
    else
        return 1;
    return 2; // warning 225: unreachable code
}
```
, I think it would make sense to adopt the same for `switch`.

### Minimal complete verifiable example (MCVE):

See above.

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Daniel-Cortez:
Fixed in #580.

Issue #559, opened 2020-08-06
Unreachable code after `goto` is not being detected


Daniel-Cortez:
### Issue description:

Currently the compiler is not able to detect unreachable code after `goto` being used on a previously defined label.
```Pawn
main()
{
    new var = 1;
label_1:
    goto label_1;
    return var; // no warning(s), even though the code on this line is not reachable
}
```
NOTE: There might also be more complicated cases, e.g.:
```Pawn
main()
{
    new const bool:TRUE = true;
    if (TRUE)
        goto label_2;
label_1:
    goto label_1;
label_2:
    return 1;
}
```
In this example the code after the `goto label_1;` line actually can be reached, which means that just having a `goto` is not enough to consider the subsequent code unreachable.
But still, I think the algorithm for reachability testing should be simple: just traverse through the table of local symbols and make sure all labels are implemented (have the `uDEFINE` flag set); if there are no unimplemented labels, then the code after `goto` is unreachable.

### Minimal complete verifiable example (MCVE):
See above.

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Daniel-Cortez:
Fixed in #580.

PR #580, opened 2020-09-30
Improvements for detection of unreachable code

### LINKED ISSUE #555: Code is falsely reported as unreachable
### Issue description:

If the `true` branch of an `if` statement ends with `return`, `break`, `continue` or an endless loop, and the `false` branch is present, but empty, the compiler thinks that the code after the `if` statement is unreachable.

```Pawn
main()
{
    new const bool:TRUE = true;
    if (TRUE) { return 1; }
    else {}
    return 0; // warning 225: unreachable code
}
```

See above.

### Workspace Information:

* Operating System:

Daniel-Cortez:

This PR fixes various bugs as well as adds new functionality for more accurate detection of unreachable code (warning 225):
* Fixes code being falsely reported as unreachable when the `true` branch of a preceding `if` statement ends with `return`, `break`, `continue` or an endless loop, and the `false` branch is empty (#555).
* Fixes unreachable code not being detected when the two `if` branches end with different kinds of terminal statements (#556).
  (By "terminal statements" I mean `return`, `break`, `continue` or an endless loop - is that a correct term for them? If not, please feel free to correct me.)
* Fixes labels being falsely detected as unreachable code, when a label is used via `goto` and then defined after an `if` statement that has both branches ending with `return` (#565).
* Adds detection of unreachable code after `goto`, `switch` and `exit` statements (#559, #560, #566).

Fixes #555, #556, #559, #560, #565, #566

* [x] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/tests/unreachable_code.pwn
+++ b/source/compiler/tests/unreachable_code.pwn
@@ -131,6 +131,20 @@ label_1:
 	// shouldn't cause warning 209 ("function should return a value")
 }
 
+test_goto_4()
+{
+	new var = 0;
+	if (var == 0)
+		goto label_1;
+	if (var != 0)
+		return 0;
+	else
+		return 1;
+	// shouldn't cause warning 225 ("unreachable code")
+label_1:
+	return 2;
+}
+
 main()
 {
 	test_if_1();
@@ -145,4 +159,5 @@ main()
 	test_goto_1();
 	test_goto_2();
 	test_goto_3();
+	test_goto_4();
 }
--- a/source/compiler/tests/unreachable_code.pwn
+++ b/source/compiler/tests/unreachable_code.pwn
@@ -25,6 +25,24 @@ test_if_2()
 	}
 }
 
+test_if_3()
+{
+	if (g_var != 0)
+		return 0;
+	else
+		return 1;
+	// shouldn't cause warning 209 ("function should return a value")
+}
+
+test_if_4()
+{
+	if (g_var != 0)
+		return 0;
+	else
+		for (;;) {}
+	// shouldn't cause warning 209 ("function should return a value")
+}
+
 test_switch_1()
 {
 	switch (g_var)
@@ -69,11 +87,22 @@ test_switch_3()
 	return 2;
 }
 
+test_endless()
+{
+	if (g_var != 0)
+		return 0;
+	for (;;) {}
+	// shouldn't cause warning 209 ("function should return a value")
+}
+
 main()
 {
 	test_if_1();
 	test_if_2();
+	test_if_3();
+	test_if_4();
 	test_switch_1();
 	test_switch_2();
 	test_switch_3();
+	test_endless();
 }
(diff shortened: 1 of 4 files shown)


source/compiler/tests/unreachable_code.pwn
```diff
@@ -0,0 +1,187 @@
...
+	// shouldn't cause warning 225 ("unreachable code")
+	return 3;
+}
+
+test_endless()
+{
+	if (g_var != 0)
+		return 0;
+	for (;;) {}
+	// shouldn't cause warning 209 ("function should return a value")
```
Y-Less, line 120: I'm wondering if this should be a different warning.  The code sometimes returns, sometimes doesn't.  The cases where you want a function that doesn't end are extremely rare IMHO, especially compared to the cases where you accidentally wrote an infinite loop and want the compiler to alert you to this fact.

Y-Less: Same for any infinite loop anywhere outside `main` TBH.

Daniel-Cortez: Well, I disabled warning 209 for infinite loops, as usually the compiler warns about them because of the loop condition having a constant result (`warning 206: redundant test: constant expression is non-zero`).
But yes, as you noticed, when a loop condition section of a `for` loop is empty and there are no `break`/`return`/`exit`/`goto` statements inside the loop body, the compiler doesn't warn about the loop being endless.
I've opened a separate issue for this in #587.

@Y-Less, if this is the only problem with this PR, can you approve it for now (and merge it now, if possible)? As I stated in #587, I'll try to implement a new warning for infinite loops, but I'll need this PR to be merged first, so later I'll be able to update its tests for the new warning.

Y-Less: Yes.  I'll merge things in a few days, as it does take a while to test things, but that's the only questionable thing I've seen.

Y-Less:
Assertion failed: sym!=NULL, file C:\Users\Alex\Documents\Work\pawn\source\compiler\sc1.c, line 5626

When I try compile my YSI test mode.

Y-Less:
OK, MCVE:

```pawn
Func()
{
	return var[_:0];
}
```

That's it.  Just that code alone with this PR will crash the compiler.

Daniel-Cortez:
Hmm, that's strange. The assertion in question is located inside function `compound()`, where normally the compiler is supposed to meet labels, not tag overrides. It's as if something forced an end of expression after `var[`, so the remaining part was treated as a separate statement, which might be a completely different bug.

Y-Less:
Yeah, I noticed that, so I double-checked the tag override was needed.  Didn't dig in to why though.

Daniel-Cortez:
Just as I thought, the compiler doesn't parse the array index sub-expression when the array symbol is invalid (#594), then treats the said sub-expression as a separate statement.

Meanwhile, I fixed the assertion in `compound()`, now the code for checking the usage of labels should work as expected.

PR #582, opened 2020-10-02
Fix false-positive warning 240 messages

Daniel-Cortez:

Fixes false-positive `warning 240: previously assigned value is never used` messages printed for local variables in (rather rare) cases when a function first assigns a value to a static variable, then calls itself recursively and uses the assigned value in the nested call.
Example:
```Pawn
RecursiveFunc()
{
    static n = 0;
    if (n == 0) // assignment "n = 1" is used here in a nested call
    {
        n = 1;
        RecursiveFunc();
        n = 0; // warning 240: previously assigned value is never used (symbol "n")
    }
}
```
There's no way to analyze such complicated uses of static local variables accurately because the compiler works in a linear way, so as a solution I disabled tracking of unused assigned values for static variables completely.

Fixes #

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -3208,7 +3208,7 @@ SC_FUNC void markinitialized(symbol *sym,int assignment)
     return;
   if (sc_status==statFIRST && (sym->vclass==sLOCAL || sym->vclass==sSTATIC))
     return;
-  if (assignment && sym->vclass!=sGLOBAL && (sym->ident==iVARIABLE || sym->ident==iREFERENCE)) {
+  if (assignment && sym->vclass==sLOCAL && (sym->ident==iVARIABLE || sym->ident==iREFERENCE)) {
     sym->usage |= uASSIGNED;
     sym->assignlevel=pc_nestlevel;
   } /* if */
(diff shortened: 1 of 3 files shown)


Y-Less (approved):

Daniel-Cortez:
Accidentally removed the branch for this PR (and by doing that closed the PR itself) when cleaning up the old branches, because I already used the same branch for #480 and #533. Should have used a different branch for each PR, I guess 😅
Anyway, the PR has been restored and there should be no changes from the moment it was approved.

Y-Less:
So there's actually another bug with static locals, even when not recursive:

https://github.com/pawn-lang/YSI-Includes/blob/5.x/YSI_Visual/y_commands/y_commands_impl.inc#L463

That line gets a warning for `sRet`, despite the fact that it might be used here:

https://github.com/pawn-lang/YSI-Includes/blob/5.x/YSI_Visual/y_commands/y_commands_impl.inc#L485

This fix does fix that warning, but it's not really the same thing and "fix" is a bit of a stretch, so I'm reporting it anyway.

PR #586, opened 2020-10-13
Branch multiline

### LINKED ISSUE #330: Multi-line strings

* [ ] Bug Report
* [x] Feature Request
* [ ] Question

The issue name is kind of self-explanatory, I suggest adding multi-line strings. These could be useful in multiple situations, I've listed a few of them myself down below in the `Usages` section.

## Syntax
I suggest using the same syntax Go uses for multi-line strings, Go uses back-ticks. Alternatively, triple double quotes are also an option `"""string content"""` (Used by Python), or maybe even just single quotes `'string content'`.

However, personally I think the Go syntax would be best.

## Usages
Mysql Queries : 
```SQL
new Cache:cache = mysql_query(MySQL,
    `SELECT
        groups.name,
        groups.points
    FROM
        groups INNER JOIN user_groups ON groups.id = user_groups.group_id
    WHERE user_groups.user_id = 1;`
);
```

Dialogs (Although in this example, it doesn't make too much of a difference) :
```C
ShowPlayerDialog(playerid, dialogid, style, caption[],
    `Dialog Line #1\n
     Dialog Line #2\n
     Dialog Line #3\n
     Dialog Line #4\n
     Dialog Line #5\n
     Dialog Line #6\n`,
button1[], button2[]);
```
Feel free to add your own usages down below as I'm sure there are a lot more.

Y-Less:

Mutli line string literals.

Fixes #330

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -940,6 +940,7 @@ SC_VDECL char sc_ctrlchar;    /* the control character (or escape character) */
 SC_VDECL char sc_ctrlchar_org;/* the default control character */
 SC_VDECL int litidx;          /* index to literal table */
 SC_VDECL int litmax;          /* current size of the literal table */
+SC_VDECL int litgrow;         /* amount to increase the literal table by */
 SC_VDECL int stgidx;          /* index to the staging buffer */
 SC_VDECL int sc_labnum;       /* number of (internal) labels */
 SC_VDECL int staging;         /* true if staging output */
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -925,6 +925,7 @@ static void initglobals(void)
   skipinput=0;           /* number of lines to skip from the first input file */
   sc_ctrlchar=CTRL_CHAR; /* the escape character */
   litmax=sDEF_LITMAX;    /* current size of the literal table */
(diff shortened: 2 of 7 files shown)


Southclaws (approved):

Y-Less:
So this broke YSI horrifically.  I know that's my problem, I just need to think of a way to solve it.  YSI currently uses \` a lot because it was a handy ASCII symbol not in use by the compiler (as is `$`, also used by YSI).  The trick is also used by several other of my libraries.  I knew this would be an issue when I wrote it, but meh.

Maybe I should switch to \`\`\` instead, for better backwards-compatibility, the major downside to that is posting code elsewhere, like github or anywhere else that uses markdown.

Y-Less:
Or...  C++ uses `R"(   )"` for multiline strings.  I'm not doing that, it's silly, but maybe something akin to the existing `!" "` and `\" "` strings.  My first thought was `^" "` but that breaks `ctrlchar` examples.   `." "` would conflict with optional parameters (not in an insurmountable way though).

Or, just make all strings always multiline.  Is there a reason there needs to be a separate syntax for it?

PR #593, opened 2020-10-20
Branch multiline

Southclaws:
Multi-line string implementation


Y-Less:
The tests fail.  I don't know why.  The output is:

```
1>
1>Expected errors:
1>
1>raw_error.pwn(1): fatal error 111: user error: "raw error\"
1>
1>Actual errors:
1>
1>raw_error.pwn(1): fatal error 111: user error: "raw error\"
```

I guess it is something to do with the escape sequence, but that's what I'm testing so I can't change that.

Y-Less:
Finally figured out why the test was failing...  I missed "Compilation aborted." in the example output, but never realised it was part of the comparison text due to the way it was laid out.

PR #598, opened 2020-10-26
Fix redefinition of enum elements

### LINKED ISSUE #597: No errors when redefining elements of the same enum
### Issue description:

Normally the compiler issues error 021 when the user tries to define two constants with the same name inside the same enum:
```Pawn
enum Enum1
{
	e0,
	e0 // error 021: symbol already defined: "e0"
};
```
But if there's another constant with the same name defined inside another named enum, for some reason the compiler doesn't print any errors, thus allowing redefinition:
```Pawn
enum Enum0 { e0 };
enum Enum1
{
	e0,
	e0 // no error(s)
};
```

See above.

### Workspace Information:

* Compiler version: 3.10.4 - 3.10.10 (interestingly, earlier versions are not affected)
* Operating System:

Daniel-Cortez:

Makes function `symbol_cache_add()` add symbols with the same name into the head of the entry list. This allows `find_symbol()` to pick the last defined symbol when there are multiple symbols with the same name, thus returning it the old behavior it had before the adoption of the hashtable-based symbol search algorithm, and fixing a problem with `findglb()` returning wrong symbols when being called from `add_constant()` (see #597).

Fixes #597

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -2850,9 +2850,8 @@ static void symbol_cache_add(symbol *sym)
     return;
   } /* if */
   cache_sym=*pcache_sym;
-  while (cache_sym->htnext!=NULL)
-    cache_sym=cache_sym->htnext;
-  cache_sym->htnext=sym;
+  *pcache_sym=sym;
+  sym->htnext=cache_sym;
 }
 
 static void symbol_cache_remove(symbol *sym)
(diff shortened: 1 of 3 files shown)


Daniel-Cortez:
Updated with a much more simple fix that doesn't introduce any extra parameters for function `findglb()` and just returns the old behavior for function `find_symbol()` (see the 1'st post for details).

PR #601, opened 2020-11-02
Fix invalid code generation for comma-separated expressions

### LINKED ISSUE #96: Inaccurate optimisation of triadic calls.
Example:

``` pawn
new x = y ? (func(), 7) : 9;
```

I would expect this code, if `y` was true, to call `func()`, execute any side-effects it has, then assign `7` to `x`, just like this (with `func()` potentially being a non-value-returning function):

``` pawn
new x;
if (y)
{
    func();
    x = 7;
}
else
{
    x = 9;
}
```

However, it seems that the compiler actually completely removes the call to `func()` as if it was never needed.  If that call does important global state modifying things, that's bad.

Daniel-Cortez:

Fixes the bug with incorrect code generation for comma-separated expressions, when the code for the whole expression gets erased if the last comma-separated sub-expression has a constant result (see https://github.com/pawn-lang/compiler/issues/96#issuecomment-403371141).

Fixes #96

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -64,6 +64,8 @@ static int constant(value *lval);
 static char lastsymbol[sNAMEMAX+1]; /* name of last function/variable */
 static int bitwise_opercount;   /* count of bitwise operators in an expression */
 static int decl_heap=0;
+static int lval_stgidx=0;
+static cell lval_cidx=0;
 
 /* Function addresses of binary operators for signed operations */
 static void (*op1[17])(void) = {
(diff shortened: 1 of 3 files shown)


Daniel-Cortez:
So Travis CI integration works now, but apparently Linux and OS X builds fail because of mixed definitions of `sc_tokens` (`extern char *sc_tokens[]` vs. `extern const char *sc_tokens[]`).
This problem comes from the changes that are already in the `dev` branch, so I'm going to fix it in a separate PR.

PR #602, opened 2020-11-03
Misc. minor fixes and adjustments

Daniel-Cortez:

* Fixes build errors on Linux and OS X (see https://github.com/pawn-lang/compiler/pull/601#issuecomment-720541010) caused by mixed declarations of array `sc_tokens` (`extern char *sc_tokens[]` vs. `extern const char *sc_tokens[]`) by moving the array to `scvars.c` and replacing all of its definitions in several source files by a single definition in `sc.h`.
* Removes the unused `uPREDEF` flag.
  After the addition of flag `flagPREDEF` (which is a direct successor to `uPREDEF`), `uPREDEF` is only checked in `delete_symbols()`, but never set.
* Removes function `finddepend()`.
  After the addition of the `symbol::child` field in 73d4934d61d0825c32faa461e9e9e7bbc99cbf38, `finddepend()` became redundant.
* Makes operator `__pragma` use a more reliable method of detecting packed strings.
The previous method didn't allow to detect empty packed strings, because internally `!""` is the same as `""`.
* Other fixes and adjustments that don't deserve a separate PR.

Fixes #

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:


Y-Less:
I merged a load, but couldn't merge this one.  There were a load of merge conflicts, which while I'm sure I could have solved individually, I can't be sure wouldn't have broken all the work.  Like some new places introduced where some of these changes should be but aren't.

Daniel-Cortez:
No problem, I'll update it.
**EDIT:** Done.

PR #537, opened 2020-06-20
[MERGED] Fix `++` and `--` not marking the symbol as "written"

Daniel-Cortez:
**What this PR does / why we need it**:

Fixes an oversight with increment and decrement operations (both pre- and post-) not marking the symbol as `uWRITTEN`.

**Which issue(s) this PR fixes**:

Fixes #536

**What kind of pull this is**:

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Some repository meta (documentation, etc)
* [ ] Other

**Additional Documentation**:

PR #613, opened 2020-11-12
Fixes for `#pragma warning`

### LINKED ISSUE #610: #pragma warning
### Issue description:
Apparently `#pragma warning disable` isn't being reset at the end of each compilation pass, resulting in its effects being carried onto the subsequent compilation passes.
```Pawn
// This should result in "warning 238: meaningless combination of class
// specifiers (const reference)", but the compiler doesn't print it, because
// warning 238 is disabled later on the 1'st pass and its on/off status
// is not being reset before the next pass.
stock Func(const &arg) return arg;
#pragma warning disable 238

main(){}
```

Another problem is that if `#pragma warning push` is used more times than `#pragma warning pop` or vice versa, the compiler silently ignores this.
```Pawn
main() {}
#pragma warning pop // no errors/warnings, although there was no "#pragma warning push"
                    // (the user probably forgot to use it)
```
I think the compiler should inform the user if they forgot to use `#pragma warning push/pop`, in the same way as it already does for `#if`/`#endif`.

```
```

### Workspace Information:

* Compiler version: 3.10.10
* Operating System:

Daniel-Cortez:

* Fixes `#pragma warning disable` not being reset at the end of each compilation pass, resulting in its effects being carried onto the subsequent compilation passes (see #610).
* Makes the compiler issue an error when `#pragma warning push` is used more times than `#pragma warning pop` and vice versa (see #610).
* Changes the return value type of `pc_pushwarnings()` and `pc_popwarnings()` from `int` to `void`.
  `pc_pushwarnings()` never returns `FALSE`; in case of memory allocation failure it calls `error(103)`, which aborts execution (as 103 is a fatal error), making the line `return FALSE;` unreachable.
  Also, the values returned by both `pc_pushwarnings()` and `pc_popwarnings()` are never used.
* Covers `#pragma warning push/pop/enable/disable` with tests (previously there weren't any).

Fixes #610

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -841,6 +841,8 @@ SC_FUNC void outinstr(const char *name,emit_outval params[],int numparams);
 /* function prototypes in SC5.C */
 SC_FUNC int error(long number,...);
 SC_FUNC void errorset(int code,int line);
+SC_FUNC void warnstack_init(void);
+SC_FUNC void warnstack_cleanup(void);
 SC_FUNC int error_suggest(int number,const char *name,const char *name2,int type,int subtype);
 
 /* function prototypes in SC6.C */
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -615,8 +615,8 @@ int pc_compile(int argc, char **argv);
 int pc_addconstant(char *name,cell value,int tag);
 int pc_addtag(char *name);
 int pc_enablewarning(int number,warnmode enable);
-int pc_pushwarnings(void);
-int pc_popwarnings(void);
+void pc_pushwarnings(void);
+void pc_popwarnings(void);
 void pc_seterrorwarnings(int enable);
 int pc_geterrorwarnings(void);
 
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -613,8 +613,10 @@ int pc_compile(int argc, char *argv[])
           error(100,incfname);          /* cannot read from ... (fatal error) */
       } /* if */
     } /* if */
+    warnstack_init();
     preprocess();                       /* fetch first line */
     parse();                            /* process all input */
+    warnstack_cleanup();
     sc_parsenum++;
   } while (sc_reparse);
 
@@ -700,8 +702,10 @@ int pc_compile(int argc, char *argv[])
     else
       plungequalifiedfile(incfname);    /* parse implicit include file (again) */
   } /* if */
+  warnstack_init();
   preprocess();                         /* fetch first line */
   parse();                              /* process all input */
+  warnstack_cleanup();
   if (sc_listing)
     goto cleanup;
   /* inpf is already closed when readline() attempts to pop of a file */
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -980,6 +980,7 @@ void parsesingleoption(char *argv);
 
 static int command(void)
 {
+  static const char str_if[]="#if...";
   int tok,ret;
   cell val;
   char *str;
@@ -1024,7 +1025,7 @@ static int command(void)
     ret=CMD_IF;
     assert(iflevel>=0);
     if (iflevel==0) {
-      error(26);                /* no matching #if */
+      error(26,str_if);         /* no matching #if */
       errorset(sRESET,0);
     } else {
       /* check for earlier #else */
@@ -1077,7 +1078,7 @@ static int command(void)
   case tpENDIF:
     ret=CMD_IF;
     if (iflevel==0){
-      error(26);        /* no matching "#if" */
+      error(26,str_if);         /* no matching "#if" */
       errorset(sRESET,0);
     } else {
       clearassignments(1);
--- a/source/compiler/sc2.c
+++ b/source/compiler/sc2.c
@@ -1308,9 +1308,10 @@ static int command(void)
           if (ok) {
             if (strcmp(str,"enable")==0 || strcmp(str,"disable")==0) {
               cell val;
+              enum s_warnmode enable=(str[0]=='e') ? warnENABLE : warnDISABLE;
               do {
                 preproc_expr(&val,NULL);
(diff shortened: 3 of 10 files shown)


afwn90cj93201nixr2e1re:
Seems like u never thought deeple where's pragma warning pop could be used.

Just create some include and name it as enable all warnings. Create different include and name it blablabla, put code

#if defined RESET_WARNINGS
#pragma warning pop
#endif

It's kinda stupid to add limitations to pop at least...

@Y-Less your thoughts?

afwn90cj93201nixr2e1re:
```pawn

#define ENABLE_ALL_WARNINGS

#include "include1.p"
#include "include2.p"
enum a {
	first_el2em = 0,
	second_ele2m = 1
}

enum b {
	first_3elem = 0,
	second_e3lem = 1
}

enum {
	first_elem = 0,
	second_elem = 1
}


main() {
 
 new c[10];
 c[first_elem] = first_elem;
 func1();
  return 0;
}

```

include2.p
```pawn

#pragma warning push
#pragma warning disable 217

func1() {
	new c;
	 c = 15;
	return c;
}
#pragma warning pop
```

include1.p
```pawn

#if defined ENABLE_ALL_WARNINGS
#pragma warning pop
#endif
```

Daniel-Cortez:
> ```Pawn
> #if defined ENABLE_ALL_WARNINGS
> #pragma warning pop
> #endif
> ```

It seems that you're completely missing the point on how `#pragma warning push/pop` works.

Firstly, `#pragma warning pop` doesn't enable all warnings, it reverts them to the state previously saved with `#pragma warning push`.

Secondly, if you thought `#pragma warning pop` would simply enable all warnings if there was no matching `#pragma warning push`, then again, no, it doesn't do anything in such cases.
For example, try compiling this code with the v3.10.10 release:
```Pawn
#pragma warning disable 203
#pragma warning pop
new var;
main(){}
```
In this example the compiler wouldn't warn about `var` being unused, because warning 203 is disabled and `#pragma warning pop` doesn't actually do anything when the warnings stack is empty (no prior `#pragma warning push`).

Thirdly, usually an unmatched `pop` is a sign that the user forgot to use `#pragma warning push` somewhere, which is why I made the compiler issue an error in such cases, to inform the user about their mistake instead of silently ignoring the problem.

afwn90cj93201nixr2e1re:
It's shouldn't be an error. No, i didn't missed it. I understand how it's works.

Y-Less:
TBH I was unsure about some of these things, but `pop` means "restore the last item".  If there's no item to restore, what should happen?

afwn90cj93201nixr2e1re:
Warning... Note, nothing. Btw it's should prevent compilation entirely.

afwn90cj93201nixr2e1re:
```pawn
#if defined SAVE_WARNINGS_STATE
#pragma warning push
#endif
```

```pawn
#if defined RESTORE_WARNINGS_STATE
#pragma warning pop
#endif
```

afwn90cj93201nixr2e1re:
It's could be complicated, i saw a lot of c code with conditional pragma warnings pushes/pops.

Y-Less:
Nothing wrong with complex sets of conditional pushes/pops, as long as they match up in the end.

afwn90cj93201nixr2e1re:
I understand this, but there are cases (i guess they r exists) where they shouldn't match. And they are justified (possible), so I suggest replacing the error with a warning at least. Same with pragma packs and stuff like that, idk should we interpret mistmatches as errors or no, but it's kinda stupid to prevent compilation just coz they aren't match up, I posted pr.

afwn90cj93201nixr2e1re:
btw, could you take a look to these:

https://github.com/pawn-lang/compiler/pull/537#issuecomment-732662560
https://github.com/pawn-lang/compiler/pull/538#issuecomment-732555125

Y-Less:
If you don't want to pop, just don't push.  Disabling a warning doesn't require a push first.

afwn90cj93201nixr2e1re:
Em, im not saying i dont wanna pooop, so, also im not saying that disabling warnings need's push before.


file1:
```
#if condition blablabla
#pragma warning push
#pragma warning disable 217
#endif
condition = 0
```

file2:
```
#if condition blablabla
#pragma warning push
#pragma warning disable 216
#endif
condition = 0
```

file3:
```
#if condition blablabla
#pragma warning push
#pragma warning disable 215
#endif
condition = 0
```
...


and after each include just pushing GENERAL pop include, it's gonna produce error, coz general include doesnt know about defines inside that
```pawn
#if defined RESTORE_WARNINGS_STATE
#pragma warning pop
#endif
```

So, idk did u get what im tryna say, but whatever, ok.



btw, could you take a look to these:

https://github.com/pawn-lang/compiler/pull/537#issuecomment-732662560
https://github.com/pawn-lang/compiler/pull/538#issuecomment-732555125

Y-Less:
But that code is probably a mistake.  Maybe it should be a warning instead, but still.  The `pop` will only restore the most recent `push`, but there are three pushes and three warning disables there.  So in that code after the `pop` only warning 215 will be restored.  You'd have to do:

```pawn
#pragma warning pop
#pragma warning pop
#pragma warning pop
```

To restore all three.

afwn90cj93201nixr2e1re:
> and after each include just pushing GENERAL pop include

```
include 1
include pop
include 2
include pop
...

What's the difference between thing that u and me said?

Daniel-Cortez:
> Maybe it should be a warning instead, but still.

Well, as I already mentioned in #610, the main reason I made them as errors was because I wanted unmatched `#pragma warning push/pop` to follow the same rules that already apply to unmatched `#if`s and `#endif`s.
I wouldn't mind changing them into warnings though, but that would require adding at least 1 new warning for this, and there's already a PR that implements a new warning (614), which probably needs to be merged in first in order to avoid possible merge conflicts.

PR #614, opened 2020-11-14
Operators `__static_assert` and `__static_check`

### LINKED ISSUE #294: static assertions

Daniel-Cortez:

Implements operators `__static_assert` and `__static_check` using @YashasSamaga's implementation as a base (see #300).

The most notable differences from the original implementation are:
* `do_static_check()` reuses the standard string parsing mechanism for the optional 2'nd argument, in the same manner as operator `__pragma` does.
* Both `__static_check` and `__static_assert` return a `bool:`-tagged result.
* The use of parentheses is now mandatory.
This is because otherwise this kind of use:
```Pawn
__static_assert 1, __static_assert 2;
```
would be ambiguous, as there will be no way to tell if it should be treated as `__static_assert(1), __static_assert(2);` or `__static_assert(1, __static_assert(2));`.

Fixes #294

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

This PR is built upon the changes from #602, so you'll want to review and merge that PR first.


Daniel-Cortez:
Rebased to the latest changes from #602.

Daniel-Cortez:
Implemented a new mechanism that allows to record tokens read by function `lex()`. With this, `__static_assert` and `__static_check` can now properly display multi-line expressions in error/warning messages.

PR #617, opened 2020-11-19
Make warning 247 work for other operators

Daniel-Cortez:
This PR does the following:
* Extends warning 247 (`use of operator "%s" on a "bool:" value`) making it work with other arithmetic and comparison operators, not only with `~`.
  Examples:
```Pawn
if (a) {}      // OK
if (!a) {}     // OK
if (a || b) {} // OK
if (a ^ b) {}  // OK
if (a && b) {} // OK
if (a == b) {} // OK
if (a != b) {} // OK
if (a <= b) {} // warning 247: use of operator "<=" on "bool:" values
if (a >= b) {} // warning 247: use of operator ">=" on "bool:" values
if (a < b) {}  // warning 247: use of operator "<" on "bool:" values
if (a > b) {}  // warning 247: use of operator ">" on "bool:" values
if (~a) {}     // warning 247: use of operator "~" on a "bool:" value; did you mean to use operator "!"?
if (-a) {}     // warning 247: use of operator "-" on a "bool:" value; did you mean to use operator "!"?
if (++a) {}    // warning 247: use of operator "++" on a "bool:" value
if (a++) {}    // warning 247: use of operator "++" on a "bool:" value
if (--a) {}    // warning 247: use of operator "--" on a "bool:" value
if (a--) {}    // warning 247: use of operator "--" on a "bool:" value
if (a * b) {}  // warning 247: use of operator "*" on "bool:" values
if (a / b) {}  // warning 247: use of operator "/" on "bool:" values
if (a % b) {}  // warning 247: use of operator "%" on "bool:" values
if (a + b) {}  // warning 247: use of operator "+" on "bool:" values
if (a - b) {}  // warning 247: use of operator "-" on "bool:" values
if (a << b) {} // warning 247: use of operator "<<" on "bool:" values
if (a >> b) {} // warning 247: use of operator ">>" on "bool:" values
if (a >>> b) {}// warning 247: use of operator ">>>" on "bool:" values
```
* Fixes the compiler calling overloaded operator `++` instead of `--` (see #627).

Fixes #

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -80,15 +80,15 @@ static void (*op1[17])(void) = {
   os_le,os_ge,os_lt,os_gt,      /* hier9, index 11 */
   ob_eq,ob_ne,                  /* hier10, index 15 */
 };
-/* These two functions are defined because the functions inc() and dec() in
- * SC4.C have a different prototype than the other code generation functions.
+/* These two macros are defined because the functions inc() and dec() in SC4.C
+ * have a different prototype than the other code generation functions.
  * The arrays for user-defined functions use the function pointers for
  * identifying what kind of operation is requested; these functions must all
  * have the same prototype. As inc() and dec() are special cases already, it
- * is simplest to add two "do-nothing" functions.
+ * is simplest to cast them into "void (*)(void)".
  */
-static void user_inc(void) {}
-static void user_dec(void) {}
+#define user_inc ((void (*)(void))inc)
+#define user_dec ((void (*)(void))dec)
 
 /*
  *  Searches for a binary operator a list of operators. The list is stored in
--- a/source/compiler/sc3.c
+++ b/source/compiler/sc3.c
@@ -61,6 +61,8 @@ static int dbltest(void (*oper)(),value *lval1,value *lval2);
 static int commutative(void (*oper)());
 static int constant(value *lval);
 
+static const char str_w247unary[]="a \"bool:\" value";
+static const char str_w247binary[]="\"bool:\" values";
 static char lastsymbol[sNAMEMAX+1]; /* name of last function/variable */
 static int bitwise_opercount;   /* count of bitwise operators in an expression */
 static int decl_heap=0;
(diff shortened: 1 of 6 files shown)


Y-Less:
This would be far better just in an include, the same as `%` for `Float:`.  I don't think tag behaviour should really be hard-coded in the compiler too much (I didn't even know there was a warning for `~`).  `bool:` is a weak tag, this makes it some sort of special-case strongish tag.

Daniel-Cortez:
> this makes it some sort of special-case strongish tag

Well, `bool:` is already kind of a special tag, as it's the only built-in tag (defined with built-in constants `true` and `false`) aside from the default one (`_:`), so I figured there would be no harm in adding extra control for operations on `bool:` values. Also, just in case, I made this warning work only if the operator in question is not overloaded, though I'm not sure why anyone would want to overload operators for `bool:` in the first place (and now that I think about it, I should've probably covered this behavior with tests).

Y-Less:
Yes, I'm also not sure, because it is special in some ways (though I don't think having predefined constants is that special really on its own).

Daniel-Cortez:
So, to sum it up, do you want me to remove this diagnostic completely, or just leave it as-is (for operator `~` only), probably?

Also, I wanted to add a similar diagnostic for comlarisons of `Float:` values with `==` and `!=` being unsafe, as they don't take the epsilon value into account (there is a way to check if rational values are fixed-point or floating-point, fixed-point values shouldn't be affected by such diagnostic), but if I understand you correctly, that would be no good either?

Y-Less:
I don't know, but really no to disabling `==` and `!=` for `Float`!  Though, again, those are defined in includes not the compiler.

Daniel-Cortez:
Added a test to make sure overloaded `bool`-tagged operators don't trigger warning 247. This PR should be ready now.

Daniel-Cortez:
Just tried to build the compiler with the changes from this PR manually in `Release` mode, in order to be double sure everything works, and it seems that the test for overloaded operators (`warning_247_ovl`) fails.
```
1>Running warning_247_ovl... FAILED
1>Test warning_247_ovl failed for the following reason: Error output didn't match
1>
1>Expected errors:
1>
1>
1>
1>Actual errors:
1>
1>warning_247_ovl.pwn(37): warning 203: symbol is never used: "operator--(bool:)"
```
Interestingly, it only fails in `Release` and `MinSizeRel` modes; in `RelWithDebInfo` and `Debug` everything works correctly. Might be some undefined behavior or something.

I'll add a "[WIP]" prefix into the PR title for now, until I find the source of this bug and fix it.

**EDIT:** Seems that the default version of the compiler (3.2.3664) is also affected by this problem, which means the changes from this PR most probably have nothing to do with it.

**EDIT (2):** Created a separate issue for this: #627.

Daniel-Cortez:
The problem should be fixed now, but I'm not sure how to really test the fix, as the problem only occurs on `Release` and `MinSizeRel` builds, and CI services build it in `RelWithDebInfo` mode, which isn't affected.

PR #620, opened 2020-11-23
revert useless error invention

afwn90cj93201nixr2e1re:
* [x] Other

### DIFF
--- a/source/compiler/sc5.c
+++ b/source/compiler/sc5.c
@@ -206,7 +206,8 @@ static char *warnmsg[] = {
 /*245*/  "enum increment \"%s %d\" has no effect on zero value (symbol \"%s\")\n",
 /*246*/  "multiplication overflow in enum element declaration (symbol \"%s\")\n",
 /*247*/  "use of operator \"~\" on a \"bool:\" value always results in \"true\"\n",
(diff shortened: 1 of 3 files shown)


afwn90cj93201nixr2e1re:
https://github.com/pawn-lang/compiler/pull/613#issuecomment-732467936
tests will fail, so, fix them by urself
@Y-Less

afwn90cj93201nixr2e1re:
btw im not sure about https://github.com/pawn-lang/compiler/pull/620/files#diff-7b7b0c0f0be41efc175e6a081244d9ee16f0b98f6d9ed3a9fe328c1d1ab6dcedR442
should we remove entire condition?
Coz seems like we should just remove one func call and condition should exist.
изображение

afwn90cj93201nixr2e1re:
https://github.com/compuphase/pawn/blob/master/compiler/sc5.c

same error here.

afwn90cj93201nixr2e1re:
So, idk how to disable tests properly...

```pawn
#if defined ENABLE_ALL_WARNINGS
#pragma warning push
#pragma warning pop
#endif
```

could fix this, but it's useless since why we should take care about it?

Issue #625, opened 2020-12-24
No errors when using `stock` on a variable with a name starting with `@`


Daniel-Cortez:
### Issue description:

The compiler issues error 042 (`invalid combination of class specifiers`) when specifier `stock` is used when defining a function whose name starts with `@` (because such names imply `public`, and public functions can't be stock), but doesn't do the same for variables.

### Minimal complete verifiable example (MCVE):

```Pawn
stock @Func(){} // error 042: invalid combination of class specifiers
stock @var = 0; // no errors/warnings
```

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Issue #624, opened 2020-12-23
`static` specifier has no effect when only used in function declaration


Daniel-Cortez:
### Issue description:

Using the `static` specifier in a new-style function declaration, but omitting it in the definition
```Pawn
// file1.inc
forward static Func();
Func(){}
```
results in `static` having no effect on the function, so the latter can be called from other files
```Pawn
// main.pwn
#include "file1.inc"

main()
{
	// This compiles without errors or warnings, even though normally
	// the scope of Func() is expected to be limited to "file1.inc" only,
	// as the function was first declared as "static".
	Func();
}
```

### Minimal complete verifiable example (MCVE):

See above.

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Daniel-Cortez:
Related: #621.

Issue #623, opened 2020-12-23
Assertion failure when defining a static global variable with a name starting with `@`


Daniel-Cortez:
### Issue description:

Compiling the following code:
```Pawn
static @var = 0;
```
causes an assertion failure when the compiler is built with assertions enabled:
```
Assertion failed: !fstatic, file source\compiler\sc1.c, line 2085
```
It seems that function `declglb()` doesn't properly check if a variable is defined with a `static` specifier, and only has an assertion for this. Normally the compiler should issue `error 042: invalid combination of class specifiers` in such cases, as names starting with `@` imply `public`.

### Minimal complete verifiable example (MCVE):

See above.

### Workspace Information:

* Compiler version: 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Issue #622, opened 2020-12-22
Combination of `static` and `stock` specifiers not allowed in function declarations


Daniel-Cortez:
### Issue description:

The compiler allows to use a combination of `static` and `stock` class specifiers in function definitions and old-style declarations, but doesn't allow it in new-style declarations.

### Minimal complete verifiable example (MCVE):

```Pawn
static stock Func(); // no errors
static stock Func() return 0; // no errors
forward static stock Func(); // error 010: invalid function or declaration
```

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

PR #631, opened 2021-01-07
Fixes for function and variable declarations

### LINKED ISSUE #621: Invalid code generation when function has `public` specifier only in its declaration
### Issue description:

```Pawn
#include <a_samp>

public Func();
Func()
{
	printf("Func()");
}

main()
{
	printf("main()");
	CallLocalFunction("Func", "");
}
```
This code makes function `main()` call itself recursively, causing a stack/heap collision.
The reason is that `Func()` is defined without the `public` specifier and the function is not called directly, so the compiler doesn't generate any code for it, thinking it's unused. But because `Func()` has a `public` specifier in its declaration, the compiler still creates an entry for this function in the public functions table.
As a result, since there's no code for `Func()`, the entry points at `main()` instead, which causes recursion.

See above.

### Workspace Information:

* Operating System:

Daniel-Cortez:

This PR does the following:
* Allows to use a combination of `static stock`/`stock static` in new-style function declarations (see #622).
* Fixes assertion failure when defining `static` variables with a name starting with `@` (see #623).
* Makes the compiler detect the use of specifier `static` on variables whose name starts with `@` as an invalid combination of class specifiers (see #625).
* Addresses problems with function class specifiers having an incorrect effect or no effect at all in function declarations (see #621, #624) by introducing the following rules:
  * Function definition must have all class specifiers that were introduced in all previous declarations of the function.
```Pawn
forward public Func1(); // Class specifier "public" is introduced; it will be required in the definition.
Func1(); // OK (previously introduced class specifiers are only mandatory in function definitions,
         // not declarations).
Func1(){} // Error (missing class specifier "public").

static Func2();  // Class specifier "static" is introduced.
forward stock Func2(); // Class specifier "stock" is introduced; now the definition is required to have
                       // both keywords "static" and "stock".
stock static Func2(){} // OK (both class specifiers "static" and "stock" are in place).
```
  * * After the function is defined, subsequent "forward" re-declarations can't introduce any new specifiers.
```Pawn
static stock Func(){}
forward public Func(); // Error (function wasn't defined as "public").
```
* Fixes a bug with the return value tag and array size being expected before the function class specifiers in new-style function declarations (see #635).
* Fixes a bug with keyword `__pragma` being expected twice in function/variable declarations starting with keywords `public`, `static` and `stock`.

Fixes #621, #622, #623, #624, #625, #635

* [x] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -229,6 +229,11 @@ typedef struct s_symbol {
 #define uRETNONE    0x010
 /* uASSIGNED indicates that a value assigned to the variable is not used yet */
 #define uASSIGNED   0x080
+/* function is declared/defined with the "static" specifier; this flag is being
+ * reset before the next compilation pass so "forward" declarations won't be
+ * required to have this specifier if the function was defined/re-declared with
+ * it only later */
+#define uSTATIC     0x200
 /* uLOOPVAR is set when a variable is read inside of a loop condition. This is
  * used to detect situations when a variable is used in a loop condition, but
  * not modified inside of a loop body. */
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -233,7 +233,15 @@ typedef struct s_symbol {
  * reset before the next compilation pass so "forward" declarations won't be
  * required to have this specifier if the function was defined/re-declared with
  * it only later */
-#define uSTATIC     0x200
+#define uDECLSTATIC 0x200
+ /* function is declared/defined with the "public" specifier; this flag is
+  * similar to "uPUBLIC", except that it's being reset before the next
+  * compilation pass so "forward" declarations won't be required to have this
+  * specifier if the function was defined/re-declared with it only later
+  * (the original "uPUBLIC" flag can't be reset at the end of compilation pass,
+  * as later it's used for generation of the table of public functions and the
+  * stubs for state functions) */
+#define uDECLPUBLIC 0x800
 /* uLOOPVAR is set when a variable is read inside of a loop condition. This is
  * used to detect situations when a variable is used in a loop condition, but
  * not modified inside of a loop body. */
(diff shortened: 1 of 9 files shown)


Y-Less:
Those missing specifier rules are causing YSI to stop building.  I know I usually say that breaking YSI isn't a blocker for new compiler versions because it does a lot of silly things, but I'm not entirely sure this is a silly thing.  If a forward declaration has `stock`, it seems more sensible to me that the `stock` rule is added to the function, regardless of what the definition then says.  Leads to some interesting things like:

```pawn
MAKE_STOCK(func);

func()
{
}
```

Daniel-Cortez:
> ```pawn
> MAKE_STOCK(func);
> 
> func()
> {
> }
> ```
Hmm... haven't seen such non-trivial application of keyword `stock` before (am I understanding this correctly that `MAKE_STOCK` is a macro that resolves into something like `stock <name>()`?), but I guess it can be useful. I'll see if I can relax those new rules for that specifier.

**EDIT:** OK, done. Also rebased the changes to the current `dev` so it would be easier to merge them.

Daniel-Cortez:
To clarify why I added this new requirement for class specifiers in function implementations, my main concern was that in 3.10.10 (without this PR) class specifiers either had no effect in function declarations (see #624) or could even make the compiler generate invalid code (#621), so this new requirement was supposed to point users at the parts of their code that could cause problems in compiler versions up to 3.10.10.

@Y-Less I already removed that rule for specifier `stock`, as it couldn't impact generated code (it only inhibits warning 204 in case the function is unused), but do you want me to remove it for `static` and `public` as well?

Y-Less:
I'm actually not sure what the most sensible result is here.  It seems there are equally good arguments for "all specifiers are additive", so this declares a `static stock` function:

```pawn
forward static Func();
forward stock Func();
Func()
{
}
```

"all specifiers must build", so only this is valid:

```pawn
forward static Func();
forward static stock Func();
static stock Func()
{
}
```

"only the definition has specifiers":

```pawn
forward Func();
forward Func();
static stock Func()
{
}
```

In that I don't think any one of them is obviously more "correct" than the others.  Having said that, I'd vote for the first one as it gives the most options.  Regardless of which is chosen the rule should be applied evenly to `stock`, `static`, and `public`, not just having one special-cased as you did for `stock`.

Y-Less:
There is another option - remove old-style declarations, but that's a bigger change.

Y-Less:
Or rahter that's a breaking change.

PR #649, opened 2021-05-04
Fix build errors on Linux and OS X

Daniel-Cortez:

Fixes build errors that originate from an incorrect declaration of an array of function pointers (initially I declared it as `static const void (*opers[])(void)` while I should have used `static void (*const opers[])(void)` - this was no problem for MSVC and MinGW, but caused errors when building for Linux and OS X with GCC/Clang).
Also removed a few excess trailing whitespaces while I was at it.

Fixes #

* [ ] A Bug Fix
* [ ] A New Feature
* [x] Other

**Additional Documentation**:


Y-Less:
Merged locally.

PR #650, opened 2021-05-05
Fix assertions for native functions returning arrays

### LINKED ISSUE #466: Native functions returning arrays and reparsing
Function stubs (native functions) that explicitly return an array don't seem to be handled well by the compiler. First, there are two asserts that fail when such a function is declared:

[sc2.c:3265](https://github.com/pawn-lang/compiler/blob/master/source/compiler/sc2.c#L3265)  
[sc3.c:2007](https://github.com/pawn-lang/compiler/blob/master/source/compiler/sc3.c#L2007)

In case of such a function, a global array is created with the same name as the function. The first assert fails because `curfunc` is not defined when the function has no body (in `funcstub`), and the second one fails because the additional variable is defined as `iARRAY`, while it should be `iREFARRAY`. Both these asserts fail when such a function is defined or called.

In addition to this, when a reparse if caused, the variable is (apparently) incorrectly deleted, resulting in a crash in `delete_symbols`. I haven't been able to find the issue, but perhaps [this](https://github.com/pawn-lang/compiler/blob/master/source/compiler/sc1.c#L3657) code is incorrect.

Daniel-Cortez:

Fixes two assertions triggering when a native function explicitly returning an array gets defined or called (see #466).

Fixes #466

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

Special thanks to @IllidanS4 not only for reporting the bug this PR fixes, but also for making a thorough research for the source of the said bug, which made the overall process of fixing it much easier.

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -3891,13 +3891,37 @@ static void funcstub(int fnative)
 
   needtoken(tTERM);
 
-  /* attach the array to the function symbol */
   if (numdim>0) {
-    assert(sym!=NULL);
-    sub=addvariable(symbolname,0,iARRAY,sGLOBAL,tag,dim,numdim,idxtag,0);
-    sub->parent=sym;
-    if (sym)
-      sym->child=sub;
+    if (sym->child==NULL) {
+      /* attach the array to the function symbol */
+      assert(sym!=NULL);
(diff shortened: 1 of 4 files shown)


Y-Less:
Merged locally.

PR #659, opened 2021-05-23
`const static` for declaring static constants

### LINKED ISSUE #70: static constant as Array size
const Size = 5;
new data[Size]; // Works fine

static const Size = 5;
new data[Size]; // error 008: must be a constant expression; assumed zero

Daniel-Cortez:

Introduces a combination of `const static` for declaring constants limited to the scope of the current source file. This combination is only valid for global constants, it's not allowed for locals.

Fixes #70

* [ ] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -3011,8 +3011,10 @@ static void decl_const(int vclass)
   char *str;
   int tag,exprtag;
   int symbolline;
+  int fstatic;
   symbol *sym;
 
+  fstatic=(vclass==sGLOBAL && matchtoken(tSTATIC));
   insert_docstring_separator();         /* see comment in newfunc() */
   do {
     tag=pc_addtag(NULL);
@@ -3025,8 +3027,11 @@ static void decl_const(int vclass)
     /* add_constant() checks for duplicate definitions */
     check_tagmismatch(tag,exprtag,FALSE,symbolline);
     sym=add_constant(constname,val,vclass,tag);
-    if (sym!=NULL)
+    if (sym!=NULL) {
+      if (fstatic)
+        sym->fnumber=fcurrent;
       sc_attachdocumentation(sym);/* attach any documentation to the constant */
+    } /* if */
   } while (matchtoken(',')); /* enddo */   /* more? */
   needtoken(tTERM);
 }
--- a/source/compiler/tests/const_static.inc
+++ b/source/compiler/tests/const_static.inc
@@ -1,0 +1,6 @@
+const static const_val = 0;
+
+PrintConstVal()
+{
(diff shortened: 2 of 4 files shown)


Y-Less:
There area  few changes I'd like to see for `const static`.  Currently adding `static` changes them in to immutable variables, instead of compile-time constants.  So while `const` can be used to declare arrays and don't get unused warnings, `const static` doesn't do either of those things.  Which makes no sense to me.

Daniel-Cortez:
> Currently adding `static` changes them in to immutable variables, instead of compile-time constants.

If you mean changing, for example, `const const_val = 0;` into `const static const_val = 0;`, then no, it would still be a compile-time constant.

> So while const can be used to declare arrays and don't get unused warnings,

Err... You mean **`static const`** can be used to declare arrays, not just `const`?

> const static doesn't do either of those things. Which makes no sense to me.

Why would you want to define an array with a combination of keywords meant for constants only? `const static` is not supposed to be interchangeable with `static const`. Currently (and by "currently" I mean in 3.10.10, not in this PR) anything that starts with keyword `const` is supposed to be a compile-time constant, and I don't think changing this behavior in the compiler would be a good idea. In other words, my intention is for `const static` to be more of a synonym of `static enum`, but without the `enum` syntax. And I think this makes a lot of sense.

Daniel-Cortez:
@Y-Less In case I got you wrong, can you give an example of how exactly you expect `const static` to work?
Also, a small correction: at the end of my previous post I meant `static enum`, not `const enum`.

Y-Less:
OK, well until you said that I didn't realise that `static const` and `const static` would be different.  I'm not sure about that at all, but that doesn't mean I'm against it.  I just thought this was a synonym for `static const`, which right now does declare an immutable variable, not a compile-time constant.  But it sounds like the behaviour you're describing for `const static` is what I meant in the first place.

Y-Less:
OK, I merged it anyway since `const static` was previously an error so while it might have made sense to have them mean the same thing, they didn't before.

Daniel-Cortez:
> OK, I merged it anyway since `const static` was previously an error so while it might have made sense to have them mean the same thing, they didn't before.

Yes, this was exactly my point. While `const static` is not synonymous to `static const`, it doesn't break anything.
My initial intention was to implement this new combination of specifiers as `static const const` (see https://github.com/pawn-lang/compiler/issues/70#issuecomment-750294881), but then it didn't sound right to me with the same specifier repeating two times, as well as it didn't seem very convenient to use, because of having more characters to type, which is why I thought `const static` would be a much better option.

PR #672, opened 2021-08-20
Cheaterman/open.mp fixes

Cheaterman:
Fixes assertion failure when calling strval("")
Fixes error handling for amx_printstring() (return sensible values to caller)


YashasSamaga:
@pawn-lang/compiler is the PR relevant? Do we use the maintain it here?

eoussama (approved):

PR #679, opened 2021-09-30
Misc. fixes and improvements for unused assignment detection

### LINKED ISSUE #669: Another warning 240 false positive.
### Issue description:

@Daniel-Cortez 

https://github.com/pawn-lang/YSI-Includes/blob/cb8a4cd99776b99b0c07933d0be29828c2199a8b/YSI_Storage/y_ini/y_ini_impl.inc#L990

The compiler reports this instance of `end` as never read, but it might be, depending on what is executed in the loop.  It is entirely possible to exit the loop without reassigning `end`.

There's also a different false positive here:

https://github.com/pawn-lang/YSI-Includes/blob/cb8a4cd99776b99b0c07933d0be29828c2199a8b/YSI_Storage/y_ini/y_ini_impl.inc#L1427

`handle` is assigned at the end of the `case` statement, but again this is within a loop and it can be used on later iterations.

```
```

### Workspace Information:

* Compiler version:
* Operating System:

Daniel-Cortez:

This PR does the following:
* Introduces warning 252 for operators `++`, `--`, and compound assignment operators (`+=`, `-=`, `*=` etc,)
```Pawn
Func(arg) {
    return arg++; // warning 252: variable has its value modified but never used: "arg"
}
```
* Fixes warning 240 false-positives when `goto` is used inside of a loop, on a label defined after the use, inside of the same loop (see https://github.com/pawn-lang/compiler/issues/669#issuecomment-895317894).
* Makes the compiler print warning 240 on the line of the current assignment instead of the previous one (see https://github.com/pawn-lang/compiler/issues/669#issuecomment-895835974).

Fixes #669

* [x] A Bug Fix
* [x] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc.h
+++ b/source/compiler/sc.h
@@ -239,6 +239,8 @@ typedef struct s_symbol {
  * so the compiler would know it shouldn't set the uLOOPVAR flag when the variable
  * is read inside a loop condition */
 #define uNOLOOPVAR  0x2000
+/* uMODIFIED indicates that the value of a variable was modified, but not used yet */
+#define uMODIFIED   0x4000
 
 #define flagDEPRECATED 0x01  /* symbol is deprecated (avoid use) */
 #define flagNAKED     0x10  /* function is naked */
@@ -319,7 +321,7 @@ typedef struct s_valuepair {
 typedef struct s_assigninfo {
   int lnumber;      /* line number of the first unused assignment made in one of
                      * the branches (used for error messages) */
-  short usage;      /* usage flags to memoize (currently only uASSIGNED) */
+  short usage;      /* usage flags to memoize */
 } symstate;
 
 /* macros for code generation */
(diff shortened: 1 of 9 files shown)


Y-Less:
It would be good if `#pragma unused x` could suppress this warning as well:

```pawn
REMOTE_FUNC__ static void:yrt_7(&a, &b, c, &d)
{
	//print("yrt_7 called");
	a = 8;
	b = a;
	c += d;
	d -= 4;
	#pragma unused c
}
```

I get a warning on the `c` here (which is a correct warning), but the `#pragma` doesn't silence it.  I'm actually testing that the value isn't propagated, so can't remove the assignment.

Daniel-Cortez:
@Y-Less Thanks, I forgot to reset the "modified" flag when using `#pragma unused/unread`. This should be fixed now.

PR #680, opened 2021-09-30
Fix crash from an incomplete `do-while` statement

### LINKED ISSUE #678: Crash from an incomplete `do-while` statement
### Issue description:
Specifying an incomplete `do-while` statement without the `while (...);` part makes the compiler misinterpret further code, which may cause a crash due to calling `fgets()` with an invalid file handle (`(FILE *)0xFFFFFFFF`).

```Pawn
main()
{
    do {}
}
```

### Workspace Information:

* Compiler version: 3.10.10
* Operating System:

Daniel-Cortez:

Fixes the compiler crash from an incomplete `do-while` statement without the `while (...);` part (see #678).
Function `test()` (file `sc1.c`) was saving and restoring the previous value of `sc_intest` through `PUSHSTK_I()/POPSTK_I()`, and this was preventing the compiler from stopping the compilation process when the end of file was reached before closing the `do-while` statement. Making the function save/restore the value of `sc_intest` using the *actual stack* (by saving the value into a local variable) fixes the issue.

Fixes #678

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -5972,6 +5972,7 @@ static int test(int label,int parens,int invert)
   cell cidx;
   int ident,tag;
   int endtok;
+  short save_intest;
   cell constval;
   symbol *sym;
   int localstaging=FALSE;
@@ -5985,7 +5986,7 @@ static int test(int label,int parens,int invert)
     #endif
   } /* if */
 
-  PUSHSTK_I(sc_intest);
+  save_intest=sc_intest;
   sc_intest=TRUE;
   endtok=0;
   if (parens!=TEST_PLAIN) {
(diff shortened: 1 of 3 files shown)


Y-Less:
Merged locally.

Y-Less:
I get an access violation here after merging this PR:

```c
/* pc_readsrc()
 * Reads a single line from the source file (or up to a maximum number of
 * characters if the line in the input file is too long).
 */
char *pc_readsrc(void *handle,unsigned char *target,int maxchars)
{
  return fgets((char*)target,maxchars,(FILE*)handle);
}
```

```
Exception thrown at 0x77A0FF05 (ntdll.dll) in pawncc.exe: 0xC0000005: Access violation writing location 0x000000B8.
```

I can't see why, looking at the change, but it was definitely introduced by 3e0c5ba9c0ae9ff3c86269010237c356d6bbd1fc

Y-Less:
When compiling the YSI test script:

https://github.com/pawn-lang/YSI/blob/master/gamemodes/YSI_TEST.pwn

Daniel-Cortez:
My bad, it seems I forgot to replace the use of `POPSTK_I` here:

https://github.com/Daniel-Cortez/pawn-3.10/blob/f053b83a48f81287c2871a5b00445fb102b73f86/source/compiler/sc1.c#L6020

This caused corruption of the value stack, would only happen when the control expression in `if`, `while`, `do..while`, `for`, `assert` or `state` results in a constant value, which is pretty rare as normally the compiler warns about this. In your case it was `assert(false)` in `YSI_Players\y_groups\y_groups_entry`.

Daniel-Cortez:
OK, now it should be fixed, both here and in #681. Not sure if I should add a separate test for this, as the case seems to be very specific.

PR #681, opened 2021-09-30
Remove keywords `*begin`, `*end` and `*then`

### LINKED ISSUE #611: Remove `*begin`, `*end` and `*then`
### Issue description:

* These keywords don't have any unique meaning; they only duplicate the functionality of the existing C-style curly bracket syntax.
* They can complicate diagnostics, causing counter-intuitive error messages, as shown in #512.
* They were already removed in Pawn 3.3, so we could do the same in this fork.

```
```

### Workspace Information:

* Compiler version: 3.10.10
* Operating System:

Daniel-Cortez:

Removes keywords `*begin`, `*end` and `*then`, as well as alternative syntaxes for `if`, `switch`, `for`, `while` and `do-while` statements, as explained in #611.

Fixes #611

* [ ] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:
This PR also includes the changes from #680 (they are needed to prevent a crash in one of the tests), so you'll probably want to review and merge that PR first.


Y-Less:
Merged locally.

Daniel-Cortez:
While working on this PR, I noticed one interesting detail:

```Pawn
main()
{
    new x = 0;
    do { ++x; } while x < 10; // this would compile, even though
                              // there are no parentheses around 'x < 10'
}
```
Apparently, the parentheses around the control expressions of `do..while` loops are optional. But for some reason they're only optional for `do-while` loops; in all other control flow statements (`while`, `for`, `if`, `switch`) parentheses are mandatory.
I'm not sure why there's this strange rule for `do..while` statements, there are no comments about this syntax in the compiler code, and the Language Guide doesn't seem to say anything about this either.

I wonder if this syntax is already used somewhere. Because if not, then removing it would allow to greatly simplify the code in function `test()` (`sc1.c`). This PR is about removing excess syntax anyway.

Issue #606, opened 2020-11-06
Compiler crash


BitFros7y:
### Issue description:
Compiler crashes with code bellow. I would expect it to at least give error (but i dont see why these code should not work).

Also, if you add size to myarray second dimension (or its first i have no idea, the one thats left for compiler to figure out). then compiler does not crash, but it starts dropping errors that make no sense at all.

### Minimal complete verifiable example (MCVE):
```
new errortest[24] = "Test";//Does not matter if initialized or not
new myarray[2][] =
{
	errortest,
	errortest
};
```
### Workspace Information:

* Compiler version: 3.10.10 (also compiled from dev branch, but crash remained)
* Command line arguments provided (or sampctl version): Standard sublime
```
[cmd: ['pawncc.exe', 'C:\\Users\\Ice\\Desktop\\Samp Server\\gamemodes\\Plugin\\Main.pwn', '-oC:\\Users\\Ice\\Desktop\\Samp Server\\gamemodes\\Plugin/Main', '-;+', '-(+', '-d3']]
[dir: C:\Users\Ice\Desktop\Samp Server\pawno]
```
* Operating System: Windows 7 SP1 x64  (build 7601)

Issue #547, opened 2020-06-25
Incorrect output of 3d-array strings


dev-karpov0:
### Issue description:

```pawn
new const Text[4][5][] = {
    {
        "Beer1",
        "Beer2",
        "Beer3",
        "Beer4",
        "Beer5"
    },
    {
        "Wine1",
        "Wine2",
        "Wine3",
        "Wine4",
        "Wine5"
    },
    {
        "Sprunk1",
        "Sprunk2",
        "Sprunk3",
        "Sprunk4",
        "Sprunk5"
    },
    {
        "Ciggy1",
        "Ciggy2",
        "Ciggy3",
        "Ciggy4",
        "Ciggy5"
    }
};
```

If size of third dimension is not specified In 3D array, the output will be as follows:

```
Beer1
Beer2
Beer3
Beer4
Beer5
-------
Wine1
Wine2
Wine3
Wine4
Wine5
-------
Sprunk1
1
nk2
runk3
Sprunk4
-------
4
nk5
ggy1
iggy2
Ciggy3
-------
```

Strings are broken in third and fourth cell of 3d array.

### Minimal complete verifiable example (MCVE):

The following code displays all strings in all cells:
```pawn
    new j = 0, k = 0;
    while(j < sizeof(Text))
    {
        while(k < sizeof(Text[]))
        {
            printf("%s", Text[j][k]);
            k++;
        }
        printf("-------");
        k = 0;
        j++;
    }
```

### Workspace Information:

* Compiler version: 3.2.3664, 3.10.10
* Command line arguments provided (or sampctl version):
* Operating System:

Issue #439, opened 2019-07-11
Assertion failure when the size of initialization data is less than the array size


Daniel-Cortez:
### Minimal complete verifiable example (MCVE):
Compiling the following code:
```Pawn
native print(const str[]);
main()
{
	static const city_names[3][] = { "Los Santos", "San Fierro" };
	print(city_names[1]);
}
```
causes an assertion failure:
```
Assertion failed: ld!=NULL, file c:\pawn\pawn-3.10\source\compiler\sc1.c, line 2588
```

And when the array has more than 2 dimensions:
```Pawn
native print(const str[]);
main()
{
	static const city_names[2][3][] = { { "Los Santos", "San Fierro" }, { "", "", "" } };
	print(city_names[1][1]);
}
```
it causes another one:
```
Assertion failed: strtol(ld->name,NULL,16)==d, file c:\pawn\pawn-3.10\source\compiler\sc1.c, line 2589
```
Special thanks to forum user UnO who first reported this at pro-pawn forums: http://pro-pawn.ru/showthread.php?16823

### Workspace Information:

* Compiler version: 3.10.9
* Command line arguments provided (or sampctl version):
* Operating System:

PR #683, opened 2021-09-30
Fix explicit zero array size being treated as no size

### LINKED ISSUE #668: Zero array size (need error 009)
`new arr[0] = {132131};
printf("arr[0] = %d", arr[0]);`

This code is fully functional, despite the zero size of the array, although there is an error 009 in the compiler that describes this case: "error 009: invalid array size (negative, zero or out of bounds)"

Daniel-Cortez:

Fixes a bug with the compiler mistreating `new array[0] = { ... };` as `new array[] = { ... };` (see #668).

Fixes #668

* [x] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -2982,7 +2982,7 @@ static cell needsub(int *tag,constvalue_root **enumroot)
     return 0;               /* zero size (like "char msg[]") */
 
   constexpr(&val,tag,&sym); /* get value (must be constant expression) */
-  if (val<0) {
+  if (val<=0) {
     error(9);               /* negative array size is invalid; assumed zero */
     val=0;
   } /* if */
--- a/source/compiler/tests/gh_668.meta
+++ b/source/compiler/tests/gh_668.meta
@@ -1,0 +1,7 @@
+{
+  'test_type': 'output_check',
+  'errors': """
+gh_668.pwn(1) : error 009: invalid array size (negative, zero or out of bounds)
+gh_668.pwn(5) : error 009: invalid array size (negative, zero or out of bounds)
+  """
+}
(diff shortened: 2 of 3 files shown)


YashasSamaga:
Please add negative tests like:

```
new arr[] = { 0 };
```

This is just to future-proof. It is currently not required since `needsub` will return early if it finds `]` without any constant expression in it and prevent the error message. 

I think the following will trigger two errors which is ok:
```
new a;
new arr[a] = { 0 };
```

`constexpr` will return zero in `val` if the expression is not constexpr. This will trigger two errors: error 20 and error 8.

Daniel-Cortez:
> 
> 
> Please add negative tests like:
> 
> ```
> new arr[] = { 0 };
> ```
> 
> This is just to future-proof.

Wouldn't this be excessive for this PR?

Don't get me wrong, I do agree that future-proofing is a good practice, and I can add such tests if you insist. But I also think that ideally we should make a single file for each group of tests (f.e. one file for testing the array syntax, another file for testing code generation for arrays, a third one for testing function declaration syntax, and so on), instead of having them scattered around over a hundred of files for each tiny bugfix we make.

Rewriting the tests might require a lot of work, and not only for the rewriting itself, but also for fixing newly discovered bugs that would very likely pop up in the process. Obviously, this is very far out of the scope of this PR, and I think all of this would only be justified when and if open.mp finally comes out, otherwise there would be no point doing this when there aren't many people left in the community and the userbase keeps decreasing with each day.

Also, we still have a few unsolved problems related to array initialization, such as #439, #547 and #606, so I think it's still too early for adding any future-proofing tests for arrays.

YashasSamaga (approved):

Y-Less:
Merged locally.

PR #714, opened 2022-07-08
FIX: Set actual default values in usage()

ghost:
Changes default values in usage() to default values instead of current values.
**NOTE**: I simply ran pawncc without arguments (and making sure no alias was set), used given values from there.
Looking at initglobals() this could still not be 100% accurate:
```c
  #if AMX_COMPACTMARGIN > 2
    sc_compress=TRUE;    /* compress output bytecodes */
  #else
    sc_compress=FALSE;
  #endif
```
AMX_COMPACTMARGIN hould be 64 by default, no idea what that does any futher (not that much into AMX). Seems like it should be fine though.

None opened

* [x ] A Bug Fix
* [ ] A New Feature
* [ ] Other

**Additional Documentation**:
N/A

### DIFF
--- a/source/compiler/sc1.c
+++ b/source/compiler/sc1.c
@@ -1567,11 +1567,11 @@ static void usage(void)
     pc_printf("         -A<num>  alignment in bytes of the data segment and the stack\n");
     pc_printf("         -a       output assembler code\n");
 #if AMX_COMPACTMARGIN > 2
-    pc_printf("         -C[+/-]  compact encoding for output file (default=%c)\n", sc_compress ? '+' : '-');
+    pc_printf("         -C[+/-]  compact encoding for output file (default=+)\n");
 #endif
     pc_printf("         -c<name> codepage name or number; e.g. 1252 for Windows Latin-1\n");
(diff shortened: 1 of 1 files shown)


Daniel-Cortez:
Firstly, you should've created an issue to report the problem in the first place (which, if I understand it correctly, is that command line arguments overwrite default values before the compiler prints them in function `usage()`, e.g. if we invoke the compiler as `pawncc -O0` it will print `default: -O0` for option `-O`, while it's supposed to print `-O1`).

Secondly, hardcoding those values at the place of their use is a bad practice. If we'll need to change them for one reason or another, we'll have to modify them not only in function `resetglobals()`, but also in `usage()`. A good solution should involve keeping those values in one place, e.g.:
```C
// somewhere at the top of sc1.c
#define sDEFAULT_DEBUG_LEVEL sCHKBOUNDS

// in initglobals():
sc_debug=sDEFAULT_DEBUG_LEVEL;

// in about():
pc_printf(" -d<num> debugging level (default=-d%d)\n", sDEFAULT_DEBUG_LEVEL);
```

